PR #9, opened 2016-03-22
libStorage Refactor (DADD)

akutz:
This patch is the basis for libStorage moving forward -- providing a framework more consistent with an evolutionary take on the Docker HTTP platform. Features include:

- Contextual logging
- Centralized types
- Generated server names
- Executor generation based on enforced Go interface types
- Librarification of the REX-Ray CLI
- Using `make run|run-debug|run-tls|run-tls-debug` to start a server directly from source, from the command line.


clintonskitson:
Make sure to run a `make deps` before executing the tests.

PR #22, opened 2016-04-14
Added Filtering to Volume and Snapshot GET REST calls

dvonthenen:
Added Filtering to Volume and Snapshot GET REST calls

Based on research on how others like Google, VMware, and etc pass filters into REST API calls for GETting objects, filter parameters can be added by appending query parameters on the URI to the REST call.

As an example for libstorage and retrieving volumes, making a call with the following URI will give you all volumes in datacenter durham.
```
GET http://<REST endpoint>/volumes?datacenter=durham
```

You can filter on multiple Fields and the behavior is such that all fields must be satisfied in order to be returned in the result set of the REST API call. For example if you have the following volume objects:
```
Volume1 {
   Fields: {"datacenter": "durham", "department": "finance"}
}
Volume2 {
   Fields: {"datacenter": "irvine", "department": "finance"}
}
```

And you make the following REST API call:
```
GET http://<REST endpoint>/volumes?datacenter=irvine&department=finance
```

Only Volume2 will be returned in the result set.

It should be noted that currently Fields aren't not being populated (I believe Chris is working on this) but I did test this by simulating the logic here with fake data:
https://github.com/dvonthenen/goprojects/blob/master/src/sandbox/main.go

### DIFF
--- a/api/server/router/snapshot/snapshot_routes.go
+++ b/api/server/router/snapshot/snapshot_routes.go
@@ -18,7 +18,7 @@ import (
 
 //the filtering mechanism applies a simple match, you could do something like
 //this future https://github.com/golang/appengine/blob/master/datastore/query.go
-func applyFilter(obj *types.Volume, filters map[string][]string) bool {
+func applyFilter(obj *types.Snapshot, filters map[string][]string) bool {
 	include := true
 	for key, values := range filters {
 		//fmt.Print("Filter Key: ", key, "\n")
--- a/api/server/router/snapshot/snapshot_routes.go
+++ b/api/server/router/snapshot/snapshot_routes.go
@@ -2,6 +2,7 @@ package snapshot
 
 import (
 	"net/http"
+	"strings"
 
 	"github.com/akutz/goof"
 
@@ -27,6 +28,10 @@ func (r *router) snapshots(
 		reply   apihttp.ServiceSnapshotMap = map[string]apihttp.SnapshotMap{}
 	)
 
+	//filtering is done by query parameters on the URI
+	var filters map[string][]string
+	filters = req.URL.Query()
+
 	for service := range services.StorageServices() {
 
 		run := func(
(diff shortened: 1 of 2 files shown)


dvonthenen:
Doh copy and paste error. Will resubmit.

akutz:
Hi @dvonthenen,

So if you look wayyy back at the original libStorage code, you'll see I was going down this route as well. Well, if you could look at it, it was pulled out and lost in a rebase. My point is that I did investigate this, and the reason I did not do it yet is because I want to figure out if the correct place to implement this is on libStorage or by way of passing filters into the backend storage platform. Consider executing a SQL filter on a client vs on the server -- the server-side execution is far more efficient. In our case the filter on the libStorage server is the same as a SQL filter on a client -- if we can move that processing to the storage platform (the server), great! 

Lastly, we can't just use *any* query strings as filter args as we use query strings for other things as well. Some prefix such as `filter-`, `lsf-`, or something could help distinguish it. That or just `filter=` and  then we create/use some type of filter syntax for the value of the query string `filter`.

I like the idea, but I don't want it to be so free-form as the included patch shows. It's just matching query strings against fields without any reason for doing so. 

Off the top of my head I think we need to:

1. Discuss this further
2. Investigate my suggestion of a static `filter` query string above using an arg value of a known/new syntax we can easily parse
3. Have pre-defined filters we activate via a query string

dvonthenen:
Hey @akutz I completely agree. I brought up the exact same example of SQL/WHERE clause while we were together in SF. We want to filter as we pull the data and not be done as a post processing step. The only advantage this has is if the engine providing the data does not have a filtering capability. Even then filtering on the client side isn't a good solution in terms of security.

When filtering is implemented properly, having an explicit "filter=" on the query string makes a lot of sense especially if we need to do some other operation, for example sort by.

For now, I am going to close this PR. We need to have a discussion then about if we need to deliver this or some other proper implementation in time for EMCW.

akutz:
Hi @dvonthenen,

So I'm thinking the right way to do this would be to implement a query string param named `filter` and give it a value that follows the ABNF ([rfc5234](https://tools.ietf.org/html/rfc5234)) LDAP string filter serach syntax ([rfc4515](https://tools.ietf.org/search/rfc4515)):

```
(Boolean-operator(filter)(filter)(filter)...)
```

or

```
(Boolean-operator(filter)((Boolean-operator(filter)(filter)))
```

For instance, here's the example in the PR's original description:

```
GET http://<REST endpoint>/volumes?datacenter=irvine&department=finance
```

Here's how it would be represented as an ABNF search string:

```
GET http://<REST endpoint>/volumes?filter=(%26(datacenter=irvine)(department=finance))
```

And let's say we always want to bring back additional attributes, such as the volume's owner?

```
GET http://<REST endpoint>/volumes?filter=(%26(datacenter=irvine)(department=finance));owner
```

Note that we're using URL-encoded value for an ampersand, `%26`, instead of `&` in the filter string since the `&` character is special for URLs and query strings in particular. 

Anyway, parsing the filter string should be fairly straight-forward, and this would enable drivers to either do filtering driver-side (via a new function on the `StorageDriver` interface named `SupportsFiltering() bool` or something of that nature) or whether the libStorage server should take on that responsibility.

Thoughts?

PR #25, opened 2016-04-15
WiP - Introducing a Filters Library

akutz:
This patch adds support for parsing ABNF LDAP filter strings in order to support filters via the URL as an argument of the query string argument named "filter".

@dvonthenen, what are your thoughts on this? Notice via the `filters_test.go` file how we actually parse them. I think it would work.

### DIFF
--- a/api/server/router/snapshot/snapshot_routes.go
+++ b/api/server/router/snapshot/snapshot_routes.go
@@ -9,9 +9,11 @@ import (
 	"github.com/emccode/libstorage/api/server/services"
 	"github.com/emccode/libstorage/api/types"
 	"github.com/emccode/libstorage/api/types/context"
+	"github.com/emccode/libstorage/api/types/drivers"
 	apihttp "github.com/emccode/libstorage/api/types/http"
 	apisvcs "github.com/emccode/libstorage/api/types/services"
 	"github.com/emccode/libstorage/api/utils"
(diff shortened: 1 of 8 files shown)


dvonthenen:
@akutz yea this looks like it should cover all our needs. Also since the code was lifted from go-ldap, I think that its probably gotten a considerable amount of testing behind it, but that doesn't preclude us from placing in our own unit tests for code coverage. I am ok with this implementation.

akutz:
Hi @dvonthenen,

I'd like you to take one more glance at this. I think this is generic enough to satisfy @clintonskitson's ask, but also adhere to something positive design-wise moving forward. Notice that we can now ask a `StorageDriver` if it supports filtering by virtue of the `SupportsFilters() bool` function. If a driver does support them, then the driver will receive a copy of the filter via the `Volumes` or `Snapshots` functions in order to apply the filter however the driver sees fit.

However, if a driver does *not* support filters then libStorage can easily apply the compiled filter against a list of volumes or snapshots (or other objects). This is where I have stopped short. I don't yet have the code written to take a compiled filter and apply it to a list of any objects in order to trim said list based on the filter. Care to take a stab at it?

dvonthenen:
@akutz I think I am following it. That should satisfy both query support on the storage driver and generic support for filtering should the storage platform not provide a filtering mechanism.

PR #39, opened 2016-04-22
Added snapshot functionality

clintonskitson:
This commit adds the snapshot functionality to the client. It introduces the create volume from, create from volume, remove, inspect, get, service get, and copy.


clintonskitson:
@akutz I believe the SnapshotCreate is the only one that modified the signatures. I opted to leverage the VolumeCreateRequest as it has the same parameters as SnapshotCreate since a volume is what comes out of it and you can specify those types of parameters when creating volumes from snapshots.

PR #46, opened 2016-04-25
Added cli server startup package as asynchronous

clintonskitson:
This commit introduces the ability to run the default server from
the cli package in as asynchronous. This enables an external
package to be able to use the server capability and be aware
of successful starts while managing it's own blocking process.


clintonskitson:
This PR also additionally will return the configuration of the server. This is useful to be able to start a client that leverages the same configuration.

clintonskitson:
This enables something like the following. 

```
	cfg, err := startServer()
	if err != nil {
		panic(err)
	}

	if err := startClient(cfg); err != nil {
		panic(err)
	}
```

PR #48, opened 2016-04-25
Semaphore Locking for Executor Downloads

akutz:
This patch uses a kernel-level semaphore to synchronize the process of downloading and executing the executor binary in order to prevent multiple threads and processes from munging access to the file.

### DIFF
--- a/api/tests/tests.go
+++ b/api/tests/tests.go
@@ -184,6 +184,10 @@ func (th *testHarness) run(
 
 				client, err := getClient(config)
 				assert.NoError(t, err)
+				assert.NotNil(t, client)
+				if client == nil {
+					panic(fmt.Sprintf("err=%v, client=%v", err, client))
+				}
 
 				for _, test := range tests {
 					test(config, client, t)
@@ -215,6 +219,11 @@ func (th *testHarness) run(
 
 					client, err := getClient(config)
 					assert.NoError(t, err)
+					assert.NotNil(t, client)
+
+					if client == nil {
+						panic(fmt.Sprintf("err=%v, client=%v", err, client))
+					}
 
 					test(config, client, t)
 				}(test, x, config)
@@ -292,7 +301,6 @@ func getClient(config gofig.Config) (client.Client, error) {
 			"host", config.Get("libstorage.host"),
 			"error dialing libStorage service", err)
 	}
-
 	return c, nil
 }
 
--- a/api/types/types.go
+++ b/api/types/types.go
@@ -1,0 +1,17 @@
+package types
+
+import (
+	"fmt"
+	"runtime"
+)
+
+// LSX is the default  name of the libStorage executor for the current OS.
+var LSX string
+
+func init() {
+	if runtime.GOOS == "windows" {
+		LSX = "lsx-windows.exe"
+	} else {
+		LSX = fmt.Sprintf("lsx-%s", runtime.GOOS)
+	}
+}
--- a/api/utils/semaphore/semaphore.go
+++ b/api/utils/semaphore/semaphore.go
@@ -1,0 +1,63 @@
+package semaphore
+
+import (
+	"os"
+	"time"
+)
+
+// Semaphore enables processes and threads to synchronize their actions.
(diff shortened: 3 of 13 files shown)


akutz:
Hi @clintonskitson, @oskoss, @vladimirvivien,

I apologize for being "offline" today, but I've been working to solve the issue of the broken builds. It was due to a concurrency issue when multiple clients attempted to read/write from the executor binary file at the same time. I *think* I've solved it by implementing named semaphores to synchronize access to the binary across processes embedded with one to many instances of the libStorage client. Please rebase your PRs from master and keep your fingers crossed.

akutz:
Note to self, see https://msdn.microsoft.com/en-us/library/windows/desktop/ms686946(v=vs.85).aspx when implementing this for Windows.

clintonskitson:
I just ran into an issue where I can't get tests to run anymore. It looks like all four runs are waiting on the executor lock. Re-rerunning `go test
```
DEBU[0001] waiting on executor lock                      host=tcp://127.0.0.1:39407 server=forest-lancer-st txCR=1461659078 txID=091b4371-6483-4688-401d-15ebd1d10b86
```

akutz:
Hi @clintonskitson,

That's a danger when working on system, in development, that make use of semaphores. They persist in the kernel. Half-run tests, or tests with errors, can leave things in an unusable state.

There are two ways to take care of the issue:

### Execute the Semaphore test suite
Running `go test ./api/utils/semaphore` will execute the semaphore test package. This [unlinks](http://man7.org/linux/man-pages/man3/sem_unlink.3.html) the semaphore with the same name used by the production code. Unlinking a semaphore destroys it immediately, removing it from the kernel. 

### Use an Unlink program
Unfortunately the program that lists and removes semaphores with OS X, [ipcs](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/ipcs.1.html) and [ipcrm](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/ipcrm.1.html) only list System V semaphores, not POSIX semaphores. To remove a POSIX semaphore here is a little C program I whipped up. Just copy this to a file called `unlink.c` and execute `gcc unlink.c -o unlink` to compile it:

```
#include <stdio.h>
#include <semaphore.h>
#include <errno.h>

int main(int argc, char** argv) {
	if (argc < 2) {
		printf("unlink: %s <semaphore_name>\n", argv[0]);
		return 1;
	}
	return sem_unlink(argv[1]) ? errno : 0;
}
```

Then, to prove it works, run the following:

```
$ go test ./drivers/storage/mock/tests -run TestClient
ok  	github.com/emccode/libstorage/drivers/storage/mock/tests	1.773s
```

Running the above test will leave a semaphore named `lsx-GOOS` on your system. That's usually fine, but no if it's left in an error'd state. To remove it do:

```
$ ./unlink /lsx-GOOS
```

Notice the leading `/` character for the semaphore name. This is required. In my case, it was `./unlink /lsx-darwin`. This should return `0` as the exit code. However, if you run it twice you'll receive an error as the semaphore is already unlinked. For example:

```
$ go test ./drivers/storage/mock/tests -run TestClient
ok  	github.com/emccode/libstorage/drivers/storage/mock/tests	1.773s

[0]akutz@pax:libstorage$ ./unlink /lsx-darwin

[0]akutz@pax:libstorage$ ./unlink /lsx-darwin

[22]akutz@pax:libstorage$
```

akutz:
FWIW, it's okay that a semaphore survives the program or test execution. What's not okay is that it survives with a zero value. If this is the case, the next program to wait on the semaphore will never stop blocking because nothing is posting (signalling) the semaphore that there is room to move. Alternatively, instead of unlinking the semaphore, we could simply signal it. That's probably a better way to handle this, but you also don't want to signal it without abandon. It needs to be exactly a single signal so the first thread waiting proceeds, the others still blocked, until the first thread is complete, and sends its own signal.

clintonskitson:
Ok, thanks for the detailed response.

PR #53, opened 2016-04-26
Added attach and detach functionality to client

clintonskitson:
This commit adds the attach and detach functionality to the
client and mock driver.

### DIFF
--- a/api/server/router/volume/volume.go
+++ b/api/server/router/volume/volume.go
@@ -74,6 +74,7 @@ func (r *router) initRoutes() {
 				schema.VolumeDetachRequestSchema,
 				schema.VolumeMapSchema,
 				func() interface{} { return &apihttp.VolumeDetachRequest{} }),
+			handlers.NewPostArgsHandler(),
 		).Queries("detach"),
 
 		// create a new volume
@@ -125,6 +126,7 @@ func (r *router) initRoutes() {
 				schema.VolumeAttachRequestSchema,
 				schema.VolumeSchema,
 				func() interface{} { return &apihttp.VolumeAttachRequest{} }),
+			handlers.NewPostArgsHandler(),
 		).Queries("attach"),
 
 		// detach all volumes for all services
@@ -136,6 +138,7 @@ func (r *router) initRoutes() {
 				schema.VolumeDetachRequestSchema,
 				schema.ServiceVolumeMapSchema,
 				func() interface{} { return &apihttp.VolumeDetachRequest{} }),
+			handlers.NewPostArgsHandler(),
 		).Queries("detach"),
 
 		// detach an individual volume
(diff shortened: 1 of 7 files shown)


api/server/router/volume/volume.go
```diff
@@ -44,6 +44,14 @@ func (r *router) initRoutes() {
 			handlers.NewSchemaValidator(nil, schema.ServiceVolumeMapSchema, nil),
 		),
 
+		// get all volumes from all services with attachments
+		httputils.NewGetRoute(
```
akutz: I'm not sure why this was added. It's unnecessary as the actual handler checks for the query string. Additionally, the same name cannot be used for two routes or one will overwrite the other.

api/server/router/volume/volume.go
```diff
@@ -125,6 +133,7 @@ func (r *router) initRoutes() {
 				schema.VolumeAttachRequestSchema,
 				schema.VolumeSchema,
 				func() interface{} { return &apihttp.VolumeAttachRequest{} }),
+			handlers.NewPostArgsHandler(),
```
akutz: A get route won't have post args, so why add it? If this is supposed to be an attach request, that already exists in this file.

```
		// attach an existing volume
		httputils.NewPostRoute(
			"volumeAttach",
			"/volumes/{service}/{volumeID}",
			r.volumeAttach,
			handlers.NewServiceValidator(),
			handlers.NewSchemaValidator(
				schema.VolumeAttachRequestSchema,
				schema.VolumeSchema,
				func() interface{} { return &apihttp.VolumeAttachRequest{} }),
		).Queries("attach"),
```

api/server/router/volume/volume.go
clintonskitson: Ok will remove it. Was troubleshooting and I believe the actual missing
piece was the line above the query string for post handler.

akutz: Cool. Yeah, the `volumes` method already checks the query string:

image

akutz: The post handler to which you referred was for a different route -- the actual attach/detach routes. The get routes, `volumes` and `volumesForService` shouldn't need post args handlers as they don't have post bodies. Did you mean the attach/detach routes when you mentioned the post args handler? I just want to make sure page==same.

clintonskitson:
Ok, so change is to remove the Queries param because it is the only handler for the post route right?

akutz:
I'm not sure anymore. I'm going to look at the file in full mode, sans diff, to see if I can understand the changes more accurately.

akutz:
So we need to remove this:

image

And then it's up to you on the post-args handler for any detach requests. Technically detach requests don't have any first-class args, but since you can pass optional data, you technically need the post args handler for any detach route, and currently said handler is missing from `volumesDetachForService` and `volumesDetachAll` in the PR.

akutz:
FWIW, the query string handler is a global handler, so no need to ever add that. The reason the post args handler isn't is because it HAS to be executed AFTER the schema handler since the post args handler uses reflection to set the post body's fields into the route's request object.

PR #88, opened 2016-05-02
Updated docker driver

clintonskitson:
This commit completes the docker functionality. The features of mount/unmount/create/move
are available with this commit.

### DIFF
--- a/drivers/integration/docker/docker.go
+++ b/drivers/integration/docker/docker.go
@@ -7,10 +7,11 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/akutz/gofig"
 	"github.com/akutz/goof"
-
 	"github.com/emccode/libstorage/api/registry"
 	"github.com/emccode/libstorage/api/types"
 	"github.com/emccode/libstorage/api/utils"
+	"strconv"
+	"time"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


drivers/integration/docker/docker.go
```diff
@@ -286,13 +353,18 @@ func (d *driver) fsType() string {
...
+}
+
 func configRegistration() *gofig.Registration {
 	r := gofig.NewRegistration("Docker")
 	r.Key(gofig.String, "", "ext4", "", "docker.fsType")
 	r.Key(gofig.String, "", "", "", "docker.volumeType")
 	r.Key(gofig.String, "", "", "", "docker.iops")
 	r.Key(gofig.String, "", "", "", "docker.size")
 	r.Key(gofig.String, "", "", "", "docker.availabilityZone")
+	r.Key(gofig.String, "", "/var/lib/rexray/volumes", "", "docker.mountDirPath")
```
akutz: We should set this value in the config instance from REX-Ray and not to this path inside libStorage. Don't you think?

drivers/integration/docker/docker.go
```diff
@@ -199,14 +200,61 @@ func (d *driver) Mount(
...
+	return d.subVolumeMountPath(mountPath), vol, nil
 }
 
 // Unmount will unmount the specified volume by volumeName or volumeID.
 func (d *driver) Unmount(
 	ctx types.Context,
 	volumeID, volumeName string,
 	opts types.Store) error {
+
+	log.WithFields(log.Fields{
```
akutz: Hi @clintonskitson,

This should just be `ctx.WithFields...`. The context type is a `log.FieldLogger`.

PR #91, opened 2016-05-04
Feature/add vb

clintonskitson:
This commit adds the VirtualBox support.


imports/local/imports_local.go
```diff
@@ -22,8 +22,8 @@ import (
 	//_ "github.com/emccode/libstorage/drivers/storage/openstack/client"
 	//_ "github.com/emccode/libstorage/drivers/storage/rackspace/client"
 	//_ "github.com/emccode/libstorage/drivers/storage/scaleio/client"
-	//_ "github.com/emccode/libstorage/drivers/storage/vbox/client"
 	_ "github.com/emccode/libstorage/drivers/storage/vfs/client"
+	_ "github.com/emccode/libstorage/drivers/storage/virtualbox"
```
akutz: Hi @clintonskitson,

This loads the entire virtualbox driver into the client, which is entire antithesis of why we have executors and the remote model :) Please fix this and re-run your tests to make sure that this doens't break things once removed.

PR #97, opened 2016-05-05
Updated virtualbox tests

clintonskitson:
This commit updates the govirtualbox tests. They model the similar
tests as scaleio.

### DIFF
--- a/drivers/storage/virtualbox/tests/vbox_test.go
+++ b/drivers/storage/virtualbox/tests/vbox_test.go
@@ -1,17 +1,18 @@
 package virtualbox
 
 import (
-	"os"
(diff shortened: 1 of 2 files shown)


drivers/storage/virtualbox/tests/vbox_test.go
```diff
@@ -1,17 +1,18 @@
-package virtualbox
+package scaleio
```
akutz: Hi @clintonskitson,

Please update this package to be `virtualbox`.

clintonskitson:
@akutz This one is good. I also included an update at the bottom to the returned size based on reporting from the test

PR #104, opened 2016-05-06
Updated log message on libs directory create

clintonskitson:
This changes the log message from info to debug when the
libstorage directory is created.

### DIFF
--- a/api/utils/paths/paths.go
+++ b/api/utils/paths/paths.go
@@ -252,7 +252,7 @@ func checkPerms(k fileKey, mustPerm bool) bool {
 			log.WithField("path", p).Debug("file exists")
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


akutz:
So that happens just once, the first time it occurs, and only if the directory doesn't exist. I like it at info so we know what directories the app is using. In testing we'll see it a lot due to rotating test dirs, so maybe it's more annoying in test than it would be in prod?

clintonskitson:
Was doing it with each CLI command from RR.

clintonskitson:
Here is the message `INFO[0000] making libStorage directory                   mustPerm=false path= perms=-rwxr-xr-x`.

akutz:
Hi @clintonskitson,

I'm closing this because it was handled elsewhere.

PR #107, opened 2016-05-07
Docker driver bug fixes, logging, and Docker 1.12 inspect

clintonskitson:
This commit has numerous bug fixes, greatly enhanced logging,
and includes functionality for Docker 1.12 Volume API. All
existing options have been tested for the package.

### DIFF
--- a/api/registry/registry_integration.go
+++ b/api/registry/registry_integration.go
@@ -6,7 +6,9 @@ import (
 	log "github.com/Sirupsen/logrus"
 	"github.com/akutz/gofig"
 
+	"github.com/emccode/libstorage/api/context"
 	"github.com/emccode/libstorage/api/types"
+	apiutils "github.com/emccode/libstorage/api/utils"
 )
 
 type idm struct {
@@ -35,13 +37,34 @@ func (d *idm) Init(ctx types.Context, config gofig.Config) error {
 	d.config = config
 	d.ctx = ctx
 	d.used = map[string]int{}
-	ctx.WithField("pathCache", "").Debug("checking volume path cache setting")
+
+	if d.pathCache() {
+		store := apiutils.NewStore()
+		store.Set("attachments", true)
+		_, _ = d.List(context.Background(), store)
+	}
+
+	ctx.WithField(types.ConfigIntegrationVolPathCache,
+		d.pathCache()).Info("cache empty path for unused volume setting")
+	ctx.WithField(types.ConfigIntegrationVolUnmountIgnoreUsed,
+		d.ignoreUsedCount()).Info("ignore used count on unmount setting")
+	ctx.WithField(types.ConfigIntegrationVolMountPreempt,
+		d.preempt()).Info("pre-emptive mount setting")
+	ctx.WithField(types.ConfigIntegrationVolCreateDisable,
+		d.disableCreate()).Info("volume creation disabled setting")
+	ctx.WithField(types.ConfigIntegrationVolRemoveDisable,
+		d.disableRemove()).Info("volume removal disabled setting")
+
 	return nil
 }
 
 func (d *idm) List(
 	ctx types.Context,
 	opts types.Store) ([]types.VolumeMapping, error) {
+
+	fields := log.Fields{
+		"opts": opts}
+	ctx.WithFields(fields).Debug("listing volumes")
 
 	volMaps, err := d.IntegrationDriver.List(ctx.Join(d.ctx), opts)
 	if err != nil {
@@ -73,6 +96,12 @@ func (d *idm) Inspect(
 	ctx types.Context,
 	volumeName string,
 	opts types.Store) (types.VolumeMapping, error) {
+
+	fields := log.Fields{
+		"volumeName": volumeName,
+		"opts":       opts}
+	ctx.WithFields(fields).Debug("inspecting volume")
+
 	return d.IntegrationDriver.Inspect(ctx.Join(d.ctx), volumeName, opts)
 }
 
@@ -80,6 +109,12 @@ func (d *idm) Mount(
 	ctx types.Context,
 	volumeID, volumeName string,
 	opts *types.VolumeMountOpts) (string, *types.Volume, error) {
+
+	fields := log.Fields{
+		"volumeName": volumeName,
+		"volumeID":   volumeID,
+		"opts":       opts}
+	ctx.WithFields(fields).Debug("mounting volume")
 
 	mp, vol, err := d.IntegrationDriver.Mount(
 		ctx.Join(d.ctx), volumeID, volumeName, opts)
@@ -94,6 +129,12 @@ func (d *idm) Unmount(
 	ctx types.Context,
 	volumeID, volumeName string,
 	opts types.Store) error {
+
+	fields := log.Fields{
+		"volumeName": volumeName,
+		"volumeID":   volumeID,
+		"opts":       opts}
+	ctx.WithFields(fields).Debug("unmounting volume")
 
 	if d.ignoreUsedCount() ||
 		d.resetCount(volumeName) ||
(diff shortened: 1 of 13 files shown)


clintonskitson:
Link to Docker PR for 1.12 and Inspect https://github.com/docker/docker/pull/21006#issuecomment-210188532

api/registry/registry_integration.go
```diff
@@ -32,14 +32,28 @@ func (d *idm) Init(ctx types.Context, config gofig.Config) error {
...
+		"empty path for unused volume setting")
+	ctx.WithField(types.ConfigVolUnmountIgnoreUsed, d.ignoreUsedCount()).Info(
+		"ignore used count on unmount setting")
+	ctx.WithField(types.ConfigVolMountPreempt, d.preempt()).Info(
+		"pre-emptive mount setting")
+	ctx.WithField(types.ConfigVolCreateDisable, d.disableCreate()).Info(
+		"volume creation disabled setting")
+	ctx.WithField(types.ConfigVolRemoveDisable, d.disableRemove()).Info(
+		"volume removal disabled setting")
+
```
akutz: If you have a bunch of settings in a row, wouldn't it be cleaner to do something
like this?

```
ctx.WithFields(log.Fields{
    types.ConfigVolPathCache: d.pathCache(),
    types.ConfigVolUnmountIgnoreUsed: d.ignoreUsedCount(),
    types.ConfigVolMountPreempt: d.preempt(),
    types.ConfigVolCreateDisable: d.disableCreate(),
    types.ConfigVolRemoveDisable: d.disableRemove(),
}).Info("integration driver settings")
```

Because, I have to say, textual descriptions of configuration keys have no place in logs like this. Not like this anyway. If information about a setting/key needs to be provided to a user/developer it should be either in the documentation or as an extension of the key's constant with some method `Doc() string`. 

See the `ContextKey` type and [the way it performs lookups](https://github.com/emccode/libstorage/blob/master/api/types/types_context.go#L116) for its text representations (although in my case I should be using `stringer` to generate code, but you get the point).

drivers/integration/docker/docker.go
```diff
@@ -25,8 +25,9 @@ type driver struct {
 }
 
 type volumeMapping struct {
-	Name             string `json:"Name"`
-	VolumeMountPoint string `json:"Mountpoint"`
+	Name             string                 `json:"Name"`
+	VolumeMountPoint string                 `json:"Mountpoint"`
+	VolumeStatus     map[string]interface{} `json:"Status"`
```
akutz: What is `VolumeStatus`? More specifically, I know what the word `Status` means, but why is it represented by a map?

drivers/integration/docker/docker.go
```diff
@@ -48,6 +53,22 @@ func newDriver() types.IntegrationDriver {
...
+	ctx.WithField(types.ConfigDockerIOPS, d.iops()).Info(
+		"default IOPS")
+	ctx.WithField(types.ConfigDockerSize, d.size()).Info(
+		"default size in gb")
+	ctx.WithField(types.ConfigDockerAvailabilityZone, d.availabilityZone()).Info(
+		"default availability zone")
+	ctx.WithField(types.ConfigDockerFsType, d.fsType()).Info(
+		"default filesystem type")
+	ctx.WithField(types.ConfigDockerMountDirPath, d.mountDirPath()).Info(
+		"default path for mounting volumes")
```
akutz: See [my comment above](https://github.com/emccode/libstorage/pull/107/files#r62412890) about printing settings.

clintonskitson:
```
root@vagrant-ubuntu-trusty-64:/home/vagrant# docker volume inspect test233
[
    {
        "Name": "test233",
        "Driver": "rexray",
        "Mountpoint": "",
        "Status": {
            "availabilityZone": "",
            "fields": {
                "polly.id": "virtualbox-c3bd0b9c-3b84-4f00-8348-88a9f9541205"
            },
            "iops": 0,
            "name": "test233",
            "server": "",
            "service": "virtualbox",
            "size": 1,
            "type": ""
        },
        "Labels": {}
    }
]
```

drivers/integration/docker/docker.go
```diff
@@ -391,41 +506,41 @@ func (d *driver) NetworkName(
...
-	r.Key(gofig.String, "", "", "", "docker.availabilityZone")
-	r.Key(gofig.String, "", "/var/lib/libstorage/volumes", "", "docker.mountDirPath")
-	r.Key(gofig.String, "", "/data", "", "linux.volume.rootpath")
+	r.Key(gofig.String, "", "ext4", "", ConfigDockerFsType)
+	r.Key(gofig.String, "", "", "", ConfigDockerVolumeType)
+	r.Key(gofig.String, "", "", "", ConfigDockerIOPS)
+	r.Key(gofig.String, "", "16", "", ConfigDockerSize)
+	r.Key(gofig.String, "", "", "", ConfigDockerAvailabilityZone)
+	r.Key(gofig.String, "", "/var/lib/libstorage/volumes", "", ConfigDockerMountDirPath)
+	r.Key(gofig.String, "", "/data", "", ConfigDockerLinuxVolumeRootPath)
```
akutz: Hi @clintonskitson,

So there's no need to register the old keys via Gofig unless you want them to be available as environment variables and command line flags. Simply checking them via the `Get___` functions is sufficient. Those properties **will** be parsed if in a config file, whether registered or not. The only thing registering them does is make them available as environment variables and command line flags.

api/registry/registry_integration.go
```diff
@@ -81,6 +108,13 @@ func (d *idm) Mount(
 	volumeID, volumeName string,
 	opts *types.VolumeMountOpts) (string, *types.Volume, error) {
 
+	fields := log.Fields{
+		"driverName": d.Name(),
```
akutz: There should be no reason to log the driver name. It should be set in the context.

api/registry/registry_integration.go
```diff
@@ -35,14 +35,34 @@ func (d *idm) Init(ctx types.Context, config gofig.Config) error {
...
+	ctx.WithField(types.ConfigIntegrationVolMountPreempt, d.preempt()).
+		WithField("serviceName", ctx.ServiceName()).Info(
+		"pre-emptive mount setting")
+	ctx.WithField(types.ConfigIntegrationVolCreateDisable, d.disableCreate()).
+		WithField("serviceName", ctx.ServiceName()).Info(
+		"volume creation disabled setting")
+	ctx.WithField(types.ConfigIntegrationVolRemoveDisable, d.disableRemove()).
+		WithField("serviceName", ctx.ServiceName()).Info(
+		"volume removal disabled setting")
+
```
akutz: Hi @clintonskitson,

It appears this section was not updated based on my previous comment. Regardless, there should be no reason to log the service name as it should be in the context. If it's not it should be added before the code gets here.

drivers/integration/docker/docker.go
clintonskitson: See the Docker spec.

https://github.com/docker/docker/pull/21006#issuecomment-210188532

drivers/integration/docker/docker.go
clintonskitson: These are the new keys as referenced in docker_config.go.

api/registry/registry_integration.go
clintonskitson: Fixed.

api/registry/registry_integration.go
clintonskitson: @akutz Having a hard time getting the ServiceName() to show up where it should. I am assuming this should be in the client.go file?

```
	if config.IsSet(types.ConfigService) {
		c.ctx = c.ctx.WithServiceName(config.GetString(types.ConfigService))
	}
```

Even with this set, it still doesn't seem to be making it down to the drivers. Ideas?

akutz: Hi @clintonskitson,

I've notices some variable scoping bugs in places. I'm finding them and fixing them. Using your example above:

```
 if config.IsSet(types.ConfigService) {
        ctx := ctx.WithServiceName(config.GetString(types.ConfigService))
}
```

I've seen stuff like that in places (they're my fault) and instead of updating the outer ctx with a new one, it's just creating a scoped ctx variable with the new value.

clintonskitson:
@akutz ServiceName() and DriverName() are gone with the assumption they would be filled in at a higher level.

PR #110, opened 2016-05-09
Re-added implicit create feature

clintonskitson:
This commit reintroduces the implicit create feature to docker.
It also includes new configurable option to control behavior.


akutz:
Hi @clintonskitson,

Do you want the setting/flag in here to be specific to Docker? Wouldn't this have benefit for other integration drivers? Shouldn't the flag be centralized and the functionality implemented in the registry manager's integration wrapper?

PR #112, opened 2016-05-09
Updated volume config keys to be in integration package

clintonskitson:
All of the configurations for volumes are not int the integration
package. The assumption is that new drivers would also use the
same keys since they could be global.


api/types/types_config_integration_compat.go
```diff
@@ -21,16 +21,47 @@ const (
...
+		{ConfigIgVolOpsRemoveDisable, ConfigOldIntegrationVolRemoveDisable},
+		{ConfigIgVolOpsUnmountIgnoreUsed, ConfigOldIntegrationVolUnmountIgnoreUsed},
+		{ConfigIgVolOpsPathCache, ConfigOldIntegrationVolPathCache},
+		{ConfigIgVolOpsCreateDefaultFsType, ConfigOldDockerFsType},
+		{ConfigIgVolOpsCreateDefaultType, ConfigOldDockerVolumeType},
+		{ConfigIgVolOpsCreateDefaultIOPS, ConfigOldDockerIOPS},
+		{ConfigIgVolOpsCreateDefaultSize, ConfigOldDockerSize},
+		{ConfigIgVolOpsCreateDefaultAZ, ConfigOldDockerAvailabilityZone},
+		{ConfigIgVolOpsMountPath, ConfigOldDockerMountDirPath},
+		{ConfigIgVolOpsMountRootPath, ConfigOldDockerLinuxVolumeRootPath},
```
akutz: why a two dimensional array instead of a map? Ordering?

clintonskitson: Why a map?

akutz: "Why a spoon cousin?" :) If I read the code correctly it's a series of key value pairs, right? I just think a map more syntactically defines that type of dependent relationship than a two dimensional array, I think most would agree here, but I'm not at all opposed to using an array like you did; I was just curious as to if there was a specific reason such as arrays maintaining ordering.

akutz: All said and done, either work, so I'll merge. I was k let ever curious as to why you elected to use the array.

PR #125, opened 2016-05-12
Fix for VBox MAC Address Proliferation

akutz:
This patch removes the VBox MAC address lookup love fest from all around the VBox storage driver and replaces it with a one-time lookup in the `InstanceInspect` method. @clintonskitson is assigned to test this patch.

### DIFF
--- a/drivers/storage/vbox/vbox.go
+++ b/drivers/storage/vbox/vbox.go
@@ -27,7 +27,7 @@ func registerConfig() {
 	r.Key(gofig.String, "", volumePath, "", "virtualbox.volumePath")
 	r.Key(gofig.String, "", "", "", "virtualbox.localMachineNameOrId")
 	r.Key(gofig.Bool, "", false, "", "virtualbox.tls")
-	r.Key(gofig.String, "", "SATAController", "", "virtualbox.controllerName")
+	r.Key(gofig.String, "", "SATA", "", "virtualbox.controllerName")
 	r.Key(gofig.String, "", "/dev/disk/by-id", "", "virtualbox.diskIDPath")
 	r.Key(gofig.String,
 		"", "/sys/class/scsi_host/", "", "virtualbox.scsiHostPath")
--- a/drivers/storage/vbox/storage/vbox_storage.go
+++ b/drivers/storage/vbox/storage/vbox_storage.go
@@ -49,10 +49,10 @@ func (d *driver) Init(ctx types.Context, config gofig.Config) error {
 		"tls":             d.tls(),
 		"volumePath":      d.volumePath(),
 		"controllerName":  d.controllerName(),
-		"machineNameOrId": d.machineNameID(ctx),
+		"machineNameOrId": d.machineNameID(""),
 	}
 
-	log.Info("initializing driver: ", fields)
+	ctx.Info("initializing driver: ", fields)
 	d.vbox = vboxc.New(d.username(), d.password(),
 		d.endpoint(), d.tls(), d.controllerName())
 
@@ -61,7 +61,7 @@ func (d *driver) Init(ctx types.Context, config gofig.Config) error {
 			"error logging in", err)
 	}
 
-	log.WithFields(fields).Info("storage driver initialized")
+	ctx.WithFields(fields).Info("storage driver initialized")
 	return nil
 }
 
@@ -94,39 +94,6 @@ func (d *driver) NextDeviceInfo(
 	return nil, nil
 }
 
-func getMacs(ctx types.Context) ([]string, error) {
-
-	iid := context.MustInstanceID(ctx)
-	var macs []string
-	if err := iid.UnmarshalMetadata(&macs); err != nil {
-		return nil, err
-	}
-	return macs, nil
-}
-
-// getInstanceID returns the local system's InstanceID.
-func (d *driver) getInstanceID(
-	ctx types.Context,
-	opts types.Store) (*types.InstanceID, error) {
-
-	macAddrs, err := getMacs(ctx)
-	if err != nil {
-		return nil, err
-	}
-
-	m, err := d.findMachine(ctx, d.machineNameID(ctx), macAddrs)
-	if err != nil {
-		goof.WithFieldE("metadata", macAddrs,
-			"failed to find local machine", err)
-	}
-
-	if m == nil {
-		return nil, goof.New("machine not found")
-	}
-
-	return &types.InstanceID{ID: m.ID}, nil
-}
-
 // InstanceInspect returns an instance.
 func (d *driver) InstanceInspect(
 	ctx types.Context,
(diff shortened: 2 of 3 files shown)


akutz:
Hi @clintonskitson,

Could you please validate this fix when you're able? Thank you.

akutz:
Hi @clintonskitson,

I'm putting together a test system locally, so I should be able to work this out in the next hour. FWIW, the `TestInstanceID` test fails as the expected `InstanceID` is still in the raw format. I'm updating the VBox code to expose the process by which an `InstanceID` is transformed inside the `InstanceInspect` method in order to make it callable outside the driver.

drivers/storage/vbox/vbox.go
```diff
@@ -27,7 +27,7 @@ func registerConfig() {
 	r.Key(gofig.String, "", volumePath, "", "virtualbox.volumePath")
 	r.Key(gofig.String, "", "", "", "virtualbox.localMachineNameOrId")
 	r.Key(gofig.Bool, "", false, "", "virtualbox.tls")
-	r.Key(gofig.String, "", "SATAController", "", "virtualbox.controllerName")
```
clintonskitson: Let's change this back to the existing setting. All of our examples that don't specify the SATA Controller expect this name. Since it is really an abritrary name, I wouldn't change how we default it at this point.

akutz:
Many thanks to @clintonskitson & @h0tbird for [REX-Ray issue #392](https://github.com/emccode/rexray/issues/392) as it helped me figure out how to configured VirtualBox for testing.

@clintonskitson, I'm approving this merge as all tests passed on my test system. Here is [the log](https://gist.github.com/akutz/4ec56869b6c3d0db0933bcec0e932b10) from that run in case you're interested in reviewing it.

clintonskitson:
https://github.com/akutz/libstorage/blob/1de598d5836f6928e52b132e246ed61b837b1ff9/drivers/storage/vbox/vbox.go#L21 

The `gotil.HomeDir` will never really work here. The home directory is from the underlying OS and not from where REX-Ray or libStorage client is running.

drivers/storage/vbox/vbox.go
akutz: Hi @clintonskitson,

I literally just merged this. The comment you made is no longer valid actually as I changed it to "SATA". I forget the original value, but was it "SATA"? If so, we're good.

clintonskitson:
I also tested your fix, and beside the default settings changes it was good for me.

akutz:
Hi @clintonskitson,

Regarding the `gotil.Homedir` comment, yes, I figured that out as well (unfortunately). The logging was crap to tell me why that was failing. Just that "unable to create medium." Nothing about invalid path. That's why I added some logging to make sure the path of the new devices is evident.

It appears that a patch is in order to fix some of the defaults anyway. So the homedir is one, what about the controller name?

clintonskitson:
It is totally an arbitrary name that differs based on who and what is spinning up the VMs under VirtualBox. If we change it now, then it could break assumptions from previous. And if we change it, then we are changing towards a name that may even be less standard. Seems best to lave it as it was.

akutz:
Hi @clintonskitson,

I must not have been clear; I don't remember what it was, and that's why I asked you. However, your comments regarding backwards compatibility seem to indicate the default value was something different previously. I don't want to break backwards compatibility.

In REX-Ray 0.3.3, the VirtualBox driver's default value for the `virtualbox.controllerName` property was [an empty string](https://github.com/emccode/rexray/blob/v0.3.3/drivers/storage/virtualbox/virtualbox.go#L694).

In @vladimirvivien's and your initial import of the VirtualBox driver into libStorage, the property's default value remained [set to an empty string](https://github.com/emccode/libstorage/commit/383aab930c505c450068e7e9358f03c9df52c981#diff-ea9c4e814373c9f0b65c9bace95db33eR149).

So given that that property has *never* had a default value, I fail to understand how prescribing a default value of `SATA`, the name VirtualBox assigns by default to new a new SATA controller, is a bad thing. One has to assume that all existing REX-Ray users that *don't* use the SATA controller are already specifying their controller name in the configuration file, and one also has to assume that even those who *are* using a SATA controller are still specifying the name of said controller since, as I've established above, previously there was **no** default value.

akutz:
Hi @clintonskitson,

One conceivable scenario in which backwards compatibility would be broken would be if the absence of a controller name meant it would default to the first controller. If that's the case, then of course I'll change the value back to an empty string.

akutz:
Hi @clintonskitson,

Per our Slack conversation, you're correct in that it wasn't empty previously, but that wasn't until commit b893be4, and the reason the default value for `controllerName` was set was invalid. I should know; I'm the one who set it:

```
[0]akutz@pax:libstorage$ for commitID in $(git log --oneline -S 'controllerName' | awk '{print $1}'); do printf "CommitID: %s\n" $commitID && git show $commitID | grep controllerName && echo; done
CommitID: b893be4
-  controllerName: SATAController
-	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")
+	r.Key(gofig.String, "", "SATAController", "", "virtualbox.controllerName")

CommitID: 9810f46
+func (d *driver) controllerName() string {
+	return d.config.GetString("virtualbox.controllerName")
+		"controllerName":  d.controllerName(),
+		d.endpoint(), d.tls(), d.controllerName())
+func (d *driver) controllerName() string {
+	return d.config.GetString("virtualbox.controllerName")
+  controllerName: SATAController
+	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")
-func (d *driver) controllerName() string {
-	return d.config.GetString("virtualbox.controllerName")
-  controllerName: SATAController
-		"controllerName":  d.controllerName(),
-		d.endpoint(), d.tls(), d.controllerName())
-func (d *driver) controllerName() string {
-	return d.config.GetString("virtualbox.controllerName")
-	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")

CommitID: 53a7f1c
-	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")
+func (d *driver) controllerName() string {
+	return d.config.GetString("virtualbox.controllerName")
+  controllerName: SATAController
-	controllerName string
+		"controllerName":  d.controllerName(),
-		d.endpoint, d.useTLS, d.controllerName)
+		d.endpoint(), d.tls(), d.controllerName())
+func (d *driver) controllerName() string {
+	return d.config.GetString("virtualbox.controllerName")
+	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")

CommitID: b122e77
+	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")
+	controllerName string
+		d.endpoint, d.useTLS, d.controllerName)
```

Commit b893be4 was the context refactor, and the only reason *I* set the default value of `controllerName` to `SATAController` is because that's what you were using in your VirtualBox tests. 

`controllerName` has always been an empty string until I mistakenly used the value from the VirtualBox test cases you wrote and made it the default value for the property. That's why I changed it to `SATA` as `SATAController` isn't the default name that VirtualBox gives to SATA controllers; `SATA` is. 

The default value for the `controllerName` property has always been an empty string and so users have always had to specify some value for it. Changing it to `SATA` now has no effect other than make it so users with the same version of VirtualBox I'm running do not have to specify a controller name at all anymore. 

No loss, minor gain. I call it a win.

Also, not to beat a dead horse, but it appears in REX-Ray's commit history the name has *always* been `SATA` when the name was specified via a configuration element:

```
[0]akutz@pax:rexray$ for commitID in $(git log --oneline -S 'controllerName' | awk '{print $1}'); do printf "CommitID: %s\n" $commitID && git show $commitID | grep controllerName && echo; done
CommitID: 1ea3dd1
-		d.endpoint(), d.tls(), d.controllerName())
-func (d *driver) controllerName() string {
-	cn := d.r.Config.GetString("virtualbox.controllerName")
-	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")

CommitID: eda24c3
-  controllerName: SATA
+  controllerName: SATA
+        controllerName: SATA
+        controllerName: SATA
+  controllerName: SATA
+              controllerName: SATA
+  controllerName: SATA
+  controllerName: SATA
+named `SATA`.  Otherwise the `controllerName` field must be populated
+  controllerName: name
-named `SATA`.  Otherwise the `controllerName` field must be populated
-  controllerName: name
-  controllerName: SATA
-  controllerName: SATA

CommitID: 6ed57af
+  controllerName: SATA

CommitID: 53f060e
-  controllerName: SATA
-  controllerName: SATA
+named `SATA`.  Otherwise the `controllerName` field must be populated
+  controllerName: name
+  controllerName: SATA
+  controllerName: SATA
-named `SATA`.  Otherwise the `controllerName` field must be populated
-  controllerName: name

CommitID: ade6919
CommitID: 2e82dc4
CommitID: ff00844
+  controllerName: SATA
+  controllerName: SATA

CommitID: 26e5625
+named `SATA`.  Otherwise the `controllerName` field must be populated
+  controllerName: name
+		d.endpoint(), d.tls(), d.controllerName())
+func (d *driver) controllerName() string {
+	cn := d.r.Config.GetString("virtualbox.controllerName")
+	r.Key(gofig.String, "", "", "", "virtualbox.controllerName")

```

akutz:
Hi @clintonskitson,

Even the [REX-Ray documentation](http://rexray.readthedocs.io/en/stable/user-guide/config/#example-sans-modules) refers to the value as `SATA`:

image


It occurs to me that this entire thing may be one big misunderstanding as perhaps you did not notice my response to your line note where I indicated I changed the default value from `SATAController` to `SATA`.

image

It doesn't change the fact that the value has always been an empty string, but maybe we're in violent agreement had you known that the default value for the property that's in the PR's commit  is the same value currently represented in the REX-Ray documentation.

clintonskitson:
Totally confused @akutz. I had just seen "SATAController" in the diff and then the change to "SATA". I just reference the 0.3.3 release, and "SATA" was the original default, so let's leave it there.

PR #134, opened 2016-05-15
Added custom headers to API methods

clintonskitson:
This commit re-introduces the ability to set customer header
values from the API client.

I haven't been able to fully end to end test this yet, but I believe this may be what is needed.

### DIFF
--- a/api/client/client.go
+++ b/api/client/client.go
@@ -1,9 +1,11 @@
 package client
 
 import (
-	"net/http"
-
+	"fmt"
 	"github.com/emccode/libstorage/api/types"
+	"net/http"
+	"regexp"
+	"sync"
 )
 
 // Client is the libStorage API client.
@@ -13,6 +15,8 @@ type client struct {
 	logRequests  bool
 	logResponses bool
 	serverName   string
+	headers      http.Header
+	headersRWL   *sync.RWMutex
 }
 
 // New returns a new API client.
@@ -21,7 +25,9 @@ func New(host string, transport *http.Transport) types.APIClient {
 		Client: http.Client{
 			Transport: transport,
 		},
-		host: host,
+		host:       host,
+		headers:    http.Header{},
+		headersRWL: &sync.RWMutex{},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


akutz:
Hi @clintonskitson,

I appreciate the intent here, but I'm going to close this PR. I've already considered how to handle other headers, and I've decided it will be done in one of two ways:

1. No Custom Headers - In this scenario all headers must be defined in advance in order to be something easily set and overridden via context values.

2. Custom Headers via Context Map - The context package itself will be expanded with `context.AddHeader` and `context.AddHeaderForDriver` such that the underlying implementation in the `lsc` context is to track *all* custom headers via a singleton map set in the context. This method enables the ability to add custom headers while still making the scoped to the context and not global to every request. The method current in the PR, one I designed, has too many flaws when the occasion arises when it is not desirable to send a previously assigned header.

Additionally, when we reopen this issue with solution No. 1 or No. 2 above, it will be for at least the 0.1.1 release as this is not a critical item, and the 0.1.0 release is due tomorrow.

If you are able to make a valid case as to why 1) this needs to be a part of 0.1.0 and 2) you cannot implement one of the above solutions, then please leave this issue closed. Otherwise I'm open to discussing it.

akutz:
Hi @clintonskitson,

FWIW, if you describe the use case we may already be able to solve it. If you're wanting to send data present in the context today, then just updated the `client_http.go` file in `./api/client` to pull that data out, check if it's valid, and set the desired headers. You could do that today with custom context keys for that matter. `ctx = ctx.WithValue("MyHeaderKey", "MyHeaderVal")`. Then later in the aforementioned file check for that key and do what you will with it. Ideally we do that I prescribed above, solutions 1 or 2. However, for a short-term fix, I prefer using a custom context key that is pulled out of said context in `client_http.go` and is set as a header.

akutz:
Hi @clintonskitson,

FYI, I'm working on this request using proposed solution #2. I'll ping you once it is merged. Would you please provide the code or links to the code where you were using the now deprecated `AddHeader` and `AddDriverHeader` functions so I can replicate their calls using the new method? Thank you.

PR #137, opened 2016-05-15
Added goisilon to glide

clintonskitson:
This commit adds the goisilon repo to glide. It was missed on the
initial commit.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: a40b76f83c0a061d7e78be85db4b98d3dc59dce7d6128a33353ed30fd8c028ab
-updated: 2016-05-12T14:11:07.282703504-05:00
+hash: 6d5f9dca49616fa3cd4d023bd751fdaa847beb6b7b6814801f14d0df4211746a
+updated: 2016-05-15T19:55:17.2880788Z
 imports:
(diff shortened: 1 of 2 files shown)


akutz:
Can you please do a glide up and also add the updated glide.lock file that gets generated? Thanks!

clintonskitson:
@akutz done.

akutz:
Hi @clintonskitson,

We cannot merge this as is. Somewhere in your project sources you have either a package:

1. That is *not* in the `github.com/emccode/libstorage` path with a dependency on `github.com/emccode/libstorage`.

2. Or you are depending upon some package that no longer exists in libStorage, but *does* exist in a branch in your fork such that Glide.

Either way, something is causing Glide to generate `glide.lock` with a full set of package names for `libStorage`, and that should never happen -- a `glide.lock` file generated that references the project Glide is managed *as* as a dependency.

I just looked, and it appears what is showing up in `glide.lock` is the commit of the current ip of `master`. So what likely happened is that you had not rebased your own master branch, and so when you built `glide.lock`by doing `glide up`, Glide decided it had to download libStorage into `./vendor` since you don't have the most up-to-date commit from the primary repo in your local `master.`

PR #140, opened 2016-05-16
Readded mock testing driver

clintonskitson:
This commit reintroduces the mock testing driver into the build.
There were small updates to latest interfaceID functionality.
The mock driver makes developing against libstorage easy for
testing as it represent a completely stateless and populated
state with multiple volumes.

### DIFF
--- a/drivers/storage/mock/executor/mock_executor.go
+++ b/drivers/storage/mock/executor/mock_executor.go
@@ -1,5 +1,3 @@
-// +build mock
-
 package executor
 
 import (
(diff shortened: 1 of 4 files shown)


clintonskitson:
@akutz Let's talk about this one. The VFS makes it more difficult for persons implementing tests around libstorage.

akutz:
Please explain in detail how VFS makes implementing testing more difficult. I disagree, but perhaps I'm missing something. VFS can do all that the Mock driver can do, and if it is lacking something it should be extended, and that's where effort should be expended, not by reintroducing a second non-production driver that must be updated when changes are made to the framework.

PR #142, opened 2016-05-19
Bugfix for lss-$GOOS, logging, unix sock, etc.

akutz:
This patch fixes several outstanding issues and introduces one new feature called autoEndpointMode:

  - The `lss-$GOOS` reference server implementation now ignores all other command line input if a configuration file is specified

  - The `libstorage.logging.level` value is now respected and both the client's and server's root contexts read this value in order to initialize the root contexts from which all client and server contexts derive with the configured log level.

  - Server endpoints configured to use UNIX sockets now have said sockets created in `$LIBSTORAGE_HOME/var/run/libstorage`.

  - The `libstorage.host` property will be used to create a default endpoint for the server if no endpoints are defined.

  - A new property, `libstorage.server.autoEndpointMode`, which defaults to `unix` and can also be set to `tcp`, is used to determine what type of endpoint to create as "localhost" if no endpoints are explicitly defined.

  - The `Server` interface is now returned from the server package instead of `io.Closer`. The `Server` interface can be used to get the name of the server, close the server, and also get all configured endpoint addresses for the server.

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -30,6 +30,8 @@ import (
 
 	"github.com/akutz/gofig"
 
+	"github.com/emccode/libstorage/api/types"
+
 	// imported to load routers
 	_ "github.com/emccode/libstorage/imports/routers"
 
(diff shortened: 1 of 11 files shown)


akutz:
Hi @clintonskitson,

See the Hound errors above. I'm going to fix the issue it calls out, but not for this release as it would break RR and Polly.

PR #150, opened 2016-05-23
Unix OS Driver

akutz:
This patch renames the Linux and Darwin OS drivers to the Unix OS driver. This driver is loaded for both the Linux and Darwin OSs. The Darwin implementation lacks implementations of the Mount & Format functions.


codenrhoden:
@akutz Is this something you still want to achieve?

akutz:
I moved this to the Backlog as it needs to be rebased and updated.

PR #167, opened 2016-05-31
Enabled preemption feature for current drivers

clintonskitson:
This commit fixes pre-emption and enables it for all current
drivers.


clintonskitson:
@akutz The delimeter for the localDevices was changed to `::` from `:` to accommodate for nfs paths that include `:` in the path.

akutz:
Huh. I'd prefer a single delimiter for all header values, but I'm not sure there is one that won't run into similar issues.

PR #176, opened 2016-06-07
Fix for Local Devices sans VolID

akutz:
This patch fixes issue 173 where drivers with local devices with no
volume IDs caused the remote call to fail.


clintonskitson:
Do you think a warning here would make sense when removing invalid values?

akutz:
Good idea

Issue #174, opened 2016-06-07
Volume.MountPoint should be populated after a mount operation


vladimirvivien:
Upon success, `libstorage/client.Integration().Mount()` returns a `mountPoint string`, a pointer to struct `libstorage/api/types.Volume`, and an `error`.  It would be nice if field `Volume.MountPoint()`  was pre-populated with the mount point value as well.

clintonskitson:
@vladimirvivien Are you referencing that the `MountPoint()` method is not returning the expected field? I am not clear and what you are describing here.

vladimirvivien:
@clintonskitson Yes, correct. `Volume.MountPoint()` is nil.

akutz:
@clintonskitson and I just talked about this, we're going to update the MountPoint on the client side during a Mount op to avoid having to do a second round-trip to reinspect the volume.

PR #177, opened 2016-06-07
Fix for Volume.MountPoint()

akutz:
This patch resolves issue #174 where the Volume.MountPoint() function
returns an empty string after the IntegrationDriver.Mount() function.
The real reason is because a second Volume inspection operation should
be executed to populate the mapping data, but to avoid the round-trip,
the mount point is now directly set on th volume's attachment object.

### DIFF
--- a/api/registry/registry_integration.go
+++ b/api/registry/registry_integration.go
@@ -130,6 +130,13 @@ func (d *idm) Mount(
 	if err != nil {
 		return "", nil, err
 	}
+
+	// if the volume has attachments assign the new mount point to the
+	// MountPoint field of the first attachment element
+	if len(vol.Attachments) > 0 {
+		vol.Attachments[0].MountPoint = mp
+	}
+
 	d.incCount(volumeName)
 	return mp, vol, err
 }


akutz:
True, but currently if you send the instance id and specify attachments to true you get the ones "attached to me." We currently only support one to one volume to instance mappings, so this should be fine I think, does that make sense?

clintonskitson:
Under the current conditions it makes sense when considering it a response
to a volumes attached and mapped to me.

akutz:
Yes, and if you follow the code as it is today, that appears to be the case. May I ask you to please confirm?

clintonskitson:
Problem I have there is that it looks to be the integration driver itself that occurs. Today it is ensured via the range https://github.com/akutz/libstorage/blob/02d7f2fc8f61da2e4b4efcfb13af7d1cc55739d2/drivers/integration/docker/docker.go#L248-L254

This may or may not be necessary, but point is that it is up to the integration driver to decide which attachments are sent back. I believe this would mean that blindly choosing the first attachment may be problematic.

Considering this logic is in place for now, I am good with a comment and addressing later when it comes to implementing the broader InstanceInspect queries.

akutz:
Okay. I wasn't soliciting validation of the current or future designs, rather I wanted you to verify the behavior today is consistent with the desired outcome. I'll change the design later as it becomes necessary to do so.

akutz:
The reason for my above comment re: validation is I also don't like the way attachments are handled currently, but I just pulled it over from RR. I agree it's not optimal, but prorities. I'll adjust this when I implement the mask design I previously discussed with you.

PR #213, opened 2016-07-01
Updated goscaleio package for call close bug

clintonskitson:
This commit fixes the call close logic to ensure it is only called
when there is a valid response.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 1be91a8105c46e6885db9d305fa22f7d5460a21f5c9815e308ed9cd61695c7d3
-updated: 2016-06-14T13:00:38.335910394-05:00
+hash: 794f4ee740f0eae62a42d92104f210d1a46f4727bbd4543fc3f41dde41fde517
+updated: 2016-07-01T13:54:05.201343222-07:00
 imports:
 - name: github.com/akutz/gofig
   version: 697c16916338166671910eeaccc50f21e3c10726
@@ -34,7 +34,7 @@ imports:
   subpackages:
   - api/v1
 - name: github.com/emccode/goscaleio
-  version: 53ea76f52205380ab52b9c1f4ad89321c286bb95
+  version: 9d95003c0069b1949a0fb46832870799caa5c360
   repo: https://github.com/emccode/goscaleio
   subpackages:
   - types/v1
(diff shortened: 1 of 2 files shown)


akutz:
Hi @clintonskitson,

Please fix this PR's commit and resend it. The issue is the `glide.yaml` file is updated but the project's `glide.lock` file is not. Simply run `make` on the project in order to solve the issue. Once the `glide.lock` file is updated I will merge this PR. 

FWIW, the issue is that when glide sees a `glide.lock` file the `glide.yaml` file is ignored. This is is a good thing as the `glide.lock` file is *all* the dependencies on which the project relies. Running `glide up` creates/updates `glide.lock` based on `glide.yaml`. However, `glide install` ignores `glide.yaml` entirely and only uses `glide.lock` if the latter is present. This ensures that the project is not rebuilt later using a new set of dependencies if `glide.yaml` and `glide up` causes `glide.lock` to be updated unintentionally.

clintonskitson:
Ok is this documented as an expectation for the future?

akutz:
Software should always be built and tested prior to submitting a patch that could affect the build process or cause a runtime failure. As any dependency change can do that, and since the build process will cause the `glide.lock` file to be updated, then adhering to standard best practices would ensure the proper files are updated.

akutz:
FWIW, I should add that the software should be built and tested prior to submitting a PR when it's possible and not cumbersome to do so, as is the case here. I will add a note in the documentation that the project expects people submitting PRs to do so. However, I also expect that most people already follow this rule based on my 15+ years of experience in the industry.

clintonskitson:
Where I may have skirted the rules is that when editing the sub-packages, the easiest approach is to edit the package from within the vendor directory to get immediate results. This then involves not using Glide to ensure it works. But a clean use of glide to build the project before checkin would be a good best practice.

clintonskitson:
This probably leans towards more explicit instructions or best practices for developing things like drivers.

For example (wrong or right just an example), driver development should be done under $GOPATH/src/github.com/emccode/rexray and the subpackages that exist under the vendor directory $GOPATH/src/github.com/emccode/rexray/vendor/github.com/emccode/libstorage and related API binding package $GOPATH/src/github.com/emccode/rexray/vendor/github.com/emccode/goscaleio.

Following the successful implementation of the driver, dependent packages should be committed and pushed to upstream repos and the glide file should be updated to accommodate. The vendor directory should then be cleared and a clean glide up should be performed.

OR

Another successful approach I've had is to remove the package from the `vendor/` folder and just build based on packages existing from $GOPATH/src and then pointing glide to the local filesystem path of the package.

Anyways, the conversation just highlights that developing a driver with sub-dependencies and what the best way to do so would be helpful that aligns against the build process etc.

akutz:
> But a clean use of glide to build the project before checkin would be a good best practice.

I do not expect this. I expect only someone build and test the project, not an end-to-end clean build. The provided suggestions regarding build processes will be taken into account, thank you. However, for now I must recommend against ever building or modifying code inside the vendor directory. That is a bad idea and leads to too many issues. Experienced engineers may sometimes do this in a pinch, but even they will often wish they had not due to the problems it can cause.

akutz:
Hi @clintonskitson,

Your latest commit has a problem. I know we've discussed this before, but I'm not sure where off the top of my head. I also know the problem is more likely to occur when code is modified inside the `vendor` directory, so please refrain from doing so.

The `glide.lock` file should not ever refer to the project for which the file exists. Notice in the newly updated file there is a new entry for `libstorage`. This is because `libstorage` is cloned inside of the `vendor` directory due to some references not being properly updated and Glide making up the difference by vendoring libStorage. Please correct this and resubmit to the PR. Thanks!

PR #230, opened 2016-07-14
Fix issue with attaching volumes when mounting to separate hosts

Oskoss:
When mounting a volume to separate hosts only the first host will actually get the mount. This is due to the logic we had in the mount function of the docker integration plugin. In order to mount we have to ensure the volume is attached. We do this by checking to see if the attachments is 0 and if it is we attach then mount, if not we just mount. This is correct when a volume can only be mounted to 1 host but when multiple hosts get involved, as in a NFS volume case,  mount will no longer mount since the volume is not attached. This PR fixes this issue by determining if the volume is attached to the host rather than looking plainly at the number of attachments.

### DIFF
--- a/drivers/integration/docker/docker.go
+++ b/drivers/integration/docker/docker.go
@@ -185,8 +185,14 @@ func (d *driver) Mount(
 		return "", nil, goof.New("no volume returned or created")
 	}
 
+	ma, err := d.localAttachmentOfVolume(ctx, vol)
+	if err != nil {
+		return "", nil, goof.New("problem getting local attachment of volume")
+	}
+
 	client := context.MustClient(ctx)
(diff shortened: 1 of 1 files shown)


akutz:
Hi @Oskoss,

Could you please sign the CLA? Thanks!

akutz:
Hi @cduchesne,

Could you please test this when you have an opportunity? Thanks!

Oskoss:
Hi @akutz 

All set.....by the way just moved to Austin. :+1: 

Thanks guys!

codenrhoden:
I think this got fixed in a different way when we redid the logic around attachment filtering. Looking at the code today, `Mount()` does check that the attachment is local. Going to close this.

PR #234, opened 2016-07-22
Update badge links in README.md

cduchesne:
The current README doesn't show all badges correctly. This changes the links to https from http, which fixes the problem.


akutz:
Hi @cduchesne,

This isn't the issue actually. I've run into it before. GitHub is notorious about caching the badge data. If you note some of the links in the README I have query strings with invalid data just to vary the URLs. This will fix things, but not permanently. I went to SSL because GitHub actually caches that data for shorter amounts of time.  Some of the issues are described [here](https://github.com/github/markup/issues/224).

cduchesne:
I tried using github's purge feature and the object still didn't pull correctly. GitHub is using Fastly so you should be able to force the purge via:
```
curl -X PURGE <url>
```
If you go to the http URLs, at least for the Go report, it auto forwards to an SSL site which may be causing GitHub's conversion from original image to camo.githubusercontent.com object. It is failing for all of our project pages with the current settings (http).

akutz:
Hi Chris,

GitHub is caching the content when it pulls it from the backend service. When you load GitHub pages, you're not pulling the content in the Markdown from the URL in the Markdown, but GitHub pulls that content aggressively on their end and you pull GH's cache.

I'm not familiar with the PURGE command you mentioned, but I know I've not seen it mentioned anywhere as it relates to this issue. I have changed from http to https and back before as a way to invalidate the GH cache because that changes the URL in the Markdown and forces GH to pull it again. I've since just added an ignored query string to force the URL to be updated.

-- 
-a

cduchesne:
Yes I was referring to github's caching server that picks up the links from the markdown. It's using camo.githubusercontent.com which is hosted by fastly. You can see an example of how to purge here: https://help.github.com/articles/why-do-my-images-have-strange-urls/ (at the bottom)

I was making the assumption that GitHub's platform for pulling the image wasn't pulling it correctly with the http request because it adds 1 additional redirect (the initial http to https). Using http serves no purpose as the initial request gets a redirect to the https site.

You should try the purge command next time and it will auto refresh within a second or two.

Here is the request to image delivered path taken.
```
[root@localhost ~]# curl -svo /dev/null http://goreportcard.com/badge/emccode/libstorage
> GET /badge/emccode/libstorage HTTP/1.1
< HTTP/1.1 301 Moved Permanently
< Location: https://goreportcard.com/badge/emccode/libstorage

[root@localhost~]# curl -svo /dev/null https://goreportcard.com/badge/emccode/libstorage
> GET /badge/emccode/libstorage HTTP/1.1
< HTTP/1.1 301 Moved Permanently
< Location: /badge/github.com/emccode/libstorage

[root@localhost ~]# curl -svo /dev/null https://goreportcard.com/badge/github.com/emccode/libstorage
> GET /badge/github.com/emccode/libstorage HTTP/1.1
< HTTP/1.1 307 Temporary Redirect
< Location: https://img.shields.io/badge/go_report-A+-brightgreen.svg?style=flat

[root@localhost ~]# curl -svo /dev/null https://img.shields.io/badge/go_report-A+-brightgreen.svg?style=flat
> GET /badge/go_report-A+-brightgreen.svg?style=flat HTTP/1.1
< image delivered
```

cduchesne:
If GitHub is using the Fastly varnish system to do the pull / caching, then it will only allow 3 restarts or 4 total loops through the varnish script. You can see here: https://docs.fastly.com/guides/vcl/mixing-and-matching-fastly-vcl-with-custom-vcl. It's quite possible that the additional redirect is causing an error to be returned instead of the image, and then it won't be cached correctly. It's possible the http to https redirect as added recently, and this could cause the breakage. I would just change it to the https version and purge manually whenever it isn't updating.

PR #243, opened 2016-08-19
OpenStack Storage Driver

### LINKED ISSUE #182: Add OpenStack storage driver
This issue is capturing the future place for the introduction of the OpenStack driver.

This driver will include existing functionality from REX-Ray and enhanced for.
- RackSpace support in this driver
- Awareness of Xen/KVM for different discovery mechanics
- Updated discovery logic

MatMaul:
fixes #182.

Notable changes :
- use cloud-init file or the metadata server to retrieve the instance uuid instead of dmidecode (kept as a fallback). dmidecode needs root so it is better to avoid it if possible.
- remove the xen-store-read executable dependency. Region parameter is optional so let's not try to guess it.
- same for the availability zone : guessing using the metadata server was wrong anyway since AZs for compute are different from AZs for volume.
- add support of trust id for keystone auth.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,8 @@ BUILD_TAGS :=   gofig \
 ifneq (true,$(TRAVIS))
 BUILD_TAGS +=   libstorage_storage_driver \
 				libstorage_storage_driver_vfs \
-				libstorage_storage_driver_openstack \
 				libstorage_storage_executor \
-				libstorage_storage_executor_vfs \
-				libstorage_storage_executor_openstack
+				libstorage_storage_executor_vfs
 endif
 endif
 
@@ -1144,6 +1142,13 @@ test-rbd-clean:
 
 test-vfs:
 	DRIVERS=vfs $(MAKE) ./drivers/storage/vfs/tests/vfs.test
+
+test-openstack:
+	DRIVERS=openstack $(MAKE) deps
+	DRIVERS=openstack $(MAKE) ./drivers/storage/openstack/tests/openstack.test
+
+test-openstack-clean:
+	DRIVERS=openstack $(MAKE) clean
 
 clean: $(GO_CLEAN)
 
--- a/Makefile
+++ b/Makefile
@@ -8,8 +8,10 @@ BUILD_TAGS :=   gofig \
 ifneq (true,$(TRAVIS))
 BUILD_TAGS +=   libstorage_storage_driver \
 				libstorage_storage_driver_vfs \
+				libstorage_storage_driver_openstack \
 				libstorage_storage_executor \
-				libstorage_storage_executor_vfs
+				libstorage_storage_executor_vfs \
+				libstorage_storage_executor_openstack
 endif
 endif
 
--- a/drivers/storage/openstack/executor/openstack_executor.go
+++ b/drivers/storage/openstack/executor/openstack_executor.go
@@ -51,15 +51,13 @@ func (d *driver) InstanceID(
 	ctx types.Context,
 	opts types.Store) (*types.InstanceID, error) {
 
-	uuid, err := getInstanceIDFromMetadataServer()
-	if err != nil {
-		metadataServerErr := err
-		uuid, err = getInstanceIDFromCloudInitFile()
-		if err != nil {
-			cloudInitFileErr := err
-			uuid, err = getInstanceIDWithDMIDecode()
-			if err != nil {
-				return nil, fmt.Errorf("%v ; %v ; %v", cloudInitFileErr, metadataServerErr, err)
+	uuid, metadataServerErr := getInstanceIDFromMetadataServer()
+	if metadataServerErr != nil {
+		uuid, cloudInitFileErr = getInstanceIDFromCloudInitFile()
+		if cloudInitFileErr != nil {
+			uuid, dmidecodeErr = getInstanceIDWithDMIDecode()
+			if dmidecodeErr != nil {
+				return nil, fmt.Errorf("%v ; %v ; %v", cloudInitFileErr, metadataServerErr, dmidecodeErr)
 			}
 		}
 	}
@@ -73,7 +71,7 @@ func getInstanceIDFromCloudInitFile() (string, error) {
 	const instanceIDFile = "/var/lib/cloud/data/instance-id"
 	idBytes, err := ioutil.ReadFile(instanceIDFile)
 	if err != nil {
-		return "", goof.WithError("error reading file "+instanceIDFile, err)
+		return "", goof.WithFieldE("file", instanceIDFile, "error reading file", err)
 	}
 
 	instanceID := string(idBytes)
@@ -115,9 +113,17 @@ func getInstanceIDWithDMIDecode() (string, error) {
 	cmd := exec.Command("dmidecode", "-t", "system")
 	cmdOut, err := cmd.Output()
 
-	if err != nil {
-		return "", goof.WithError("error calling dmidecode", err)
+	if exiterr, ok := err.(*exec.ExitError); ok {
+		stderr := string(exiterr.Stderr)
+		ctx.WithError(
+			exiterr,
+		).WithField(
+			"stderr", stderr,
+		).Error("error calling dmidecode")
+		return nil,
+			goof.Newf("error calling dmidecode: %s", stderr)
 	}
+	return nil, goof.WithError("error calling dmidecode", err)
 
 	rp := regexp.MustCompile("UUID:(.*)")
 	uuid := strings.Replace(rp.FindString(string(cmdOut)), "UUID: ", "", -1)
@@ -140,8 +146,7 @@ func (d *driver) LocalDevices(
 	contentBytes, err := ioutil.ReadFile(file)
 	if err != nil {
 		return nil,
-			goof.WithFieldsE(
-				goof.Fields{"file": file}, "error reading file", err)
+			goof.WithFieldE("file", file, "error reading file", err)
 	}
 
 	content := string(contentBytes)
--- a/drivers/storage/openstack/executor/openstack_executor.go
+++ b/drivers/storage/openstack/executor/openstack_executor.go
@@ -8,19 +8,26 @@ import (
 
 	"github.com/codedellemc/libstorage/api/registry"
 	"github.com/codedellemc/libstorage/api/types"
+	"github.com/codedellemc/libstorage/api/utils"
 	"github.com/codedellemc/libstorage/drivers/storage/openstack"
+	// the next import initializes and registers the OS drivers
+	_ "github.com/codedellemc/libstorage/imports/local"
 
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
+	"runtime"
 	"strings"
+	"time"
 )
 
 type driver struct {
-	config gofig.Config
+	config   gofig.Config
+	osDriver types.OSDriver
 }
 
 func init() {
@@ -33,6 +40,15 @@ func newDriver() types.StorageExecutor {
 
 func (d *driver) Init(ctx types.Context, config gofig.Config) error {
 	d.config = config
+
+	var err error
+	if d.osDriver, err = registry.NewOSDriver(runtime.GOOS); err != nil {
+		return err
+	}
+	if err = d.osDriver.Init(ctx, config); err != nil {
+		return err
+	}
+
 	return nil
 }
 
@@ -40,24 +56,18 @@ func (d *driver) Name() string {
 	return openstack.Name
 }
 
-// InstanceID returns the local instance ID for the test
-func InstanceID(config gofig.Config) (*types.InstanceID, error) {
-	d := newDriver()
-	d.Init(nil, config)
-	return d.InstanceID(nil, nil)
-}
-
-func (d *driver) InstanceID(
-	ctx types.Context,
-	opts types.Store) (*types.InstanceID, error) {
-
-	uuid, metadataServerErr := getInstanceIDFromMetadataServer()
-	if metadataServerErr != nil {
-		uuid, cloudInitFileErr = getInstanceIDFromCloudInitFile()
-		if cloudInitFileErr != nil {
-			uuid, dmidecodeErr = getInstanceIDWithDMIDecode()
-			if dmidecodeErr != nil {
-				return nil, fmt.Errorf("%v ; %v ; %v", cloudInitFileErr, metadataServerErr, dmidecodeErr)
+func (d *driver) InstanceID(ctx types.Context, opts types.Store) (*types.InstanceID, error) {
+	fields := map[string]interface{}{}
+	uuid, err := getInstanceIDFromConfigDrive(ctx, d)
+	if err != nil {
+		fields["configDrive"] = err
+		uuid, err = getInstanceIDFromMetadataServer()
+		if err != nil {
+			fields["metadataServer"] = err
+			uuid, err = getInstanceIDWithDMIDecode()
+			if err != nil {
+				fields["dmidecode"] = err
+				return nil, goof.WithFields(fields, "unable to get InstanceID from any sources")
 			}
 		}
 	}
@@ -67,33 +77,9 @@ func (d *driver) InstanceID(
 	return iid, nil
 }
 
-func getInstanceIDFromCloudInitFile() (string, error) {
-	const instanceIDFile = "/var/lib/cloud/data/instance-id"
-	idBytes, err := ioutil.ReadFile(instanceIDFile)
-	if err != nil {
-		return "", goof.WithFieldE("file", instanceIDFile, "error reading file", err)
-	}
-
-	instanceID := string(idBytes)
-	instanceID = strings.TrimSpace(instanceID)
-
-	return instanceID, nil
-}
-
-func getInstanceIDFromMetadataServer() (string, error) {
-	const metadataURL = "http://169.254.169.254/openstack/latest/meta_data.json"
-	resp, err := http.Get(metadataURL)
-	if err != nil {
-		return "", goof.WithError("error getting metadata from "+metadataURL, err)
-	}
-	defer resp.Body.Close()
-
-	data, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return "", goof.WithError("io error reading metadata", err)
-	}
+func parseUUID(metadata []byte) (string, error) {
 	var decodedJSON interface{}
-	err = json.Unmarshal(data, &decodedJSON)
+	err := json.Unmarshal(metadata, &decodedJSON)
 	if err != nil {
 		return "", goof.WithError("error unmarshalling metadata", err)
 	}
@@ -109,21 +95,97 @@ func getInstanceIDFromMetadataServer() (string, error) {
 	return uuid, nil
 }
 
-func getInstanceIDWithDMIDecode() (string, error) {
-	cmd := exec.Command("dmidecode", "-t", "system")
-	cmdOut, err := cmd.Output()
-
+func execCommand(cmd string, args ...string) (string, error) {
+	command := exec.Command(cmd, args...)
+	out, err := command.Output()
 	if exiterr, ok := err.(*exec.ExitError); ok {
 		stderr := string(exiterr.Stderr)
-		ctx.WithError(
-			exiterr,
-		).WithField(
-			"stderr", stderr,
-		).Error("error calling dmidecode")
-		return nil,
-			goof.Newf("error calling dmidecode: %s", stderr)
+		return "", goof.WithFieldE("stderr", stderr, "execute command failed", err)
+	} else if err != nil {
+		return "", goof.WithError("execute command failed", err)
 	}
-	return nil, goof.WithError("error calling dmidecode", err)
+	return string(out), nil
+}
+
+// the code of getInstanceIDFromConfigDrive is mostly copied from k8s OpenStack driver
+// https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/openstack/metadata.go
+// Copyright to the original authors (Apache license)
+
+// Config drive is defined as an iso9660 or vfat (deprecated) drive
+// with the "config-2" label.
+// http://docs.openstack.org/user-guide/cli-config-drive.html
+const configDriveLabel = "config-2"
+const configDrivePath = "openstack/latest/meta_data.json"
+
+func getInstanceIDFromConfigDrive(ctx types.Context, d *driver) (string, error) {
+	// Try to read instance UUID from config drive.
+	dev := "/dev/disk/by-label/" + configDriveLabel
+	if _, err := os.Stat(dev); os.IsNotExist(err) {
+		cmdOut, err := execCommand(
+			"blkid", "-l",
+			"-t", "LABEL="+configDriveLabel,
+			"-o", "device",
+		)
+
+		if err != nil {
+			return "", goof.WithError("Unable to run blkid", err)
+		}
+		dev = strings.TrimSpace(string(cmdOut))
+	}
+
+	mntdir, err := ioutil.TempDir("", "configdrive")
+	if err != nil {
+		return "", err
+	}
+	defer os.Remove(mntdir)
+
+	mountOpts := types.DeviceMountOpts{
+		FsType:       "iso9660",
+		MountOptions: "ro",
+	}
+	err = d.osDriver.Mount(ctx, dev, mntdir, &mountOpts)
+	if err != nil {
+		mountOpts.FsType = "vfat"
+		err = d.osDriver.Mount(ctx, dev, mntdir, &mountOpts)
+	}
+	if err != nil {
+		return "", goof.WithFieldE("device", dev, "error mounting configdrive", err)
+	}
+	defer d.osDriver.Unmount(ctx, mntdir, utils.NewStore())
+
+	metadataBytes, err := ioutil.ReadFile(filepath.Join(mntdir, configDrivePath))
+	if err != nil {
+		return "", goof.WithError("error reading metadata file on config drive", err)
+	}
+
+	return parseUUID(metadataBytes)
+}
+
+func getInstanceIDFromMetadataServer() (string, error) {
+	const metadataURL = "http://169.254.169.254/openstack/latest/meta_data.json"
+	httpClient := http.Client{
+		Timeout: time.Duration(5 * time.Second),
+	}
+
+	resp, err := httpClient.Get(metadataURL)
+	if err != nil {
+		return "", goof.WithFieldE("url", metadataURL, "error getting metadata from server", err)
+	}
+	defer resp.Body.Close()
+
+	metadataBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", goof.WithError("io error reading metadata", err)
+	}
+
+	return parseUUID(metadataBytes)
+}
+
+func getInstanceIDWithDMIDecode() (string, error) {
+	cmdOut, err := execCommand("dmidecode", "-t", "system")
+	if err != nil {
+		return "", goof.WithError("error calling dmidecode", err)
+	}
 
 	rp := regexp.MustCompile("UUID:(.*)")
 	uuid := strings.Replace(rp.FindString(string(cmdOut)), "UUID: ", "", -1)
--- a/drivers/storage/openstack/executor/openstack_executor.go
+++ b/drivers/storage/openstack/executor/openstack_executor.go
@@ -1,0 +1,153 @@
+package executor
+
+import (
+	gofig "github.com/akutz/gofig/types"
+	"github.com/akutz/goof"
+
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+	"github.com/codedellemc/libstorage/drivers/storage/openstack"
+
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os/exec"
+	"regexp"
+	"strings"
+)
+
+type driver struct {
+	config gofig.Config
+}
+
+func init() {
+	registry.RegisterStorageExecutor(openstack.Name, newDriver)
+}
+
+func newDriver() types.StorageExecutor {
+	return &driver{}
+}
+
+func (d *driver) Init(ctx types.Context, config gofig.Config) error {
+	d.config = config
+	return nil
+}
+
+func (d *driver) Name() string {
+	return openstack.Name
+}
+
+func (d *driver) InstanceID(
+	ctx types.Context,
+	opts types.Store) (*types.InstanceID, error) {
+
+	uuid, err := getInstanceIDFromMetadataServer()
+	if err != nil {
+		metadataServerErr := err
+		uuid, err = getInstanceIDFromCloudInitFile()
+		if err != nil {
+			cloudInitFileErr := err
+			uuid, err = getInstanceIDWithDMIDecode()
+			if err != nil {
+				return nil, fmt.Errorf("%v ; %v ; %v", cloudInitFileErr, metadataServerErr, err)
+			}
+		}
+	}
+
+	iid := &types.InstanceID{Driver: openstack.Name, ID: strings.ToLower(uuid)}
+
+	return iid, nil
+}
+
+func getInstanceIDFromCloudInitFile() (string, error) {
+	const instanceIDFile = "/var/lib/cloud/data/instance-id"
+	idBytes, err := ioutil.ReadFile(instanceIDFile)
+	if err != nil {
+		return "", goof.WithError("error reading file "+instanceIDFile, err)
+	}
+
+	instanceID := string(idBytes)
+	instanceID = strings.TrimSpace(instanceID)
+
+	return instanceID, nil
+}
+
+func getInstanceIDFromMetadataServer() (string, error) {
+	const metadataURL = "http://169.254.169.254/openstack/latest/meta_data.json"
+	resp, err := http.Get(metadataURL)
+	if err != nil {
+		return "", goof.WithError("error getting metadata from "+metadataURL, err)
+	}
+	defer resp.Body.Close()
+
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", goof.WithError("io error reading metadata", err)
+	}
+	var decodedJSON interface{}
+	err = json.Unmarshal(data, &decodedJSON)
+	if err != nil {
+		return "", goof.WithError("error unmarshalling metadata", err)
+	}
+	decodedJSONMap, ok := decodedJSON.(map[string]interface{})
+	if !ok {
+		return "", goof.New("error casting metadata decoded JSON")
+	}
+	uuid, ok := decodedJSONMap["uuid"].(string)
+	if !ok {
+		return "", goof.New("error casting metadata uuid field")
+	}
+
+	return uuid, nil
+}
+
+func getInstanceIDWithDMIDecode() (string, error) {
+	cmd := exec.Command("dmidecode", "-t", "system")
+	cmdOut, err := cmd.Output()
+
+	if err != nil {
+		return "", goof.WithError("error calling dmidecode", err)
+	}
+
+	rp := regexp.MustCompile("UUID:(.*)")
+	uuid := strings.Replace(rp.FindString(string(cmdOut)), "UUID: ", "", -1)
+
+	return strings.ToLower(uuid), nil
+}
+
+func (d *driver) NextDevice(
+	ctx types.Context,
+	opts types.Store) (string, error) {
+	return "", types.ErrNotImplemented
+}
+
+func (d *driver) LocalDevices(
+	ctx types.Context,
+	opts *types.LocalDevicesOpts) (*types.LocalDevices, error) {
+	devicesMap := make(map[string]string)
+
+	file := "/proc/partitions"
+	contentBytes, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil,
+			goof.WithFieldsE(
+				goof.Fields{"file": file}, "error reading file", err)
+	}
+
+	content := string(contentBytes)
+
+	lines := strings.Split(content, "\n")
+	for _, line := range lines[2:] {
+		fields := strings.Fields(line)
+		if len(fields) >= 4 {
+			devicePath := "/dev/" + fields[3]
+			devicesMap[devicePath] = ""
+		}
+	}
+
+	return &types.LocalDevices{
+		Driver:    openstack.Name,
+		DeviceMap: devicesMap,
+	}, nil
+}
--- a/drivers/storage/openstack/executor/openstack_executor.go
+++ b/drivers/storage/openstack/executor/openstack_executor.go
@@ -1,3 +1,5 @@
+// +build !libstorage_storage_executor libstorage_storage_executor_openstack
+
 package executor
 
 import (
@@ -36,6 +38,13 @@ func (d *driver) Init(ctx types.Context, config gofig.Config) error {
 
 func (d *driver) Name() string {
 	return openstack.Name
+}
+
+// InstanceID returns the local instance ID for the test
+func InstanceID(config gofig.Config) (*types.InstanceID, error) {
+	d := newDriver()
+	d.Init(nil, config)
+	return d.InstanceID(nil, nil)
 }
 
 func (d *driver) InstanceID(
--- a/drivers/storage/openstack/storage/openstack_storage.go
+++ b/drivers/storage/openstack/storage/openstack_storage.go
@@ -104,7 +104,11 @@ func (d *driver) Init(context types.Context, config gofig.Config) error {
 	fields["domainName"] = d.domainName()
 
 	trustID := d.trustID()
-	fields["trustId"] = trustID
+	if trustID == "" {
+		fields["trustId"] = ""
+	} else {
+		fields["trustId"] = "******"
+	}
 
 	d.provider, err = openstack.NewClient(authOpts.IdentityEndpoint)
 	if err != nil {
@@ -382,7 +386,7 @@ func (d *driver) VolumeSnapshot(
 
 	ctx.WithFields(fields).Info("waiting for snapshot creation to complete")
 
-	err = snapshots.WaitForStatus(d.clientBlockStorage, snapshot.ID, "available", int(d.volumeSnapshotTimeout().Seconds()))
+	err = snapshots.WaitForStatus(d.clientBlockStorage, snapshot.ID, "available", int(d.snapshotTimeout().Seconds()))
 	if err != nil {
 		return nil,
 			goof.WithFieldsE(fields,
@@ -391,6 +395,21 @@ func (d *driver) VolumeSnapshot(
 
 	return translateSnapshot(snapshot), nil
 
+}
+
+func waitFor404(c *gophercloud.ServiceClient, url string, secs int) error {
+	return gophercloud.WaitFor(secs, func() (bool, error) {
+		ret, err := c.Get(url, nil, nil)
+
+		if err != nil {
+			if ret != nil && ret.StatusCode == 404 {
+				return true, nil
+			}
+			return false, err
+		}
+
+		return false, nil
+	})
 }
 
 func (d *driver) SnapshotRemove(
@@ -402,6 +421,10 @@ func (d *driver) SnapshotRemove(
 		return goof.WithFieldE("snapshotId", snapshotID, "error removing snapshot", resp.Err)
 	}
 
+	err := waitFor404(d.clientBlockStorage, d.clientBlockStorage.ServiceURL("snapshots", snapshotID), int(d.deleteTimeout().Seconds()))
+	if err != nil {
+		return goof.WithFieldE("snapshotId", snapshotID, "error waiting for snapshot removal", err)
+	}
 	return nil
 }
 
@@ -486,7 +509,7 @@ func (d *driver) createVolume(
 		fields["volumeId"] = volume.ID
 
 		ctx.WithFields(fields).Info("waiting for volume creation to complete")
-		err = volumes.WaitForStatus(d.clientBlockStoragev2, volume.ID, "available", int(d.volumeCreateTimeout().Seconds()))
+		err = volumes.WaitForStatus(d.clientBlockStoragev2, volume.ID, "available", int(d.createTimeout().Seconds()))
 		if err != nil {
 			return nil,
 				goof.WithFieldsE(fields,
@@ -505,7 +528,7 @@ func (d *driver) createVolume(
 	fields["volumeId"] = volume.ID
 
 	ctx.WithFields(fields).Info("waiting for volume creation to complete")
-	err = volumesv1.WaitForStatus(d.clientBlockStorage, volume.ID, "available", int(d.volumeCreateTimeout().Seconds()))
+	err = volumesv1.WaitForStatus(d.clientBlockStorage, volume.ID, "available", int(d.createTimeout().Seconds()))
 	if err != nil {
 		return nil,
 			goof.WithFieldsE(fields,
@@ -526,16 +549,21 @@ func (d *driver) VolumeRemove(
 		return goof.WithFields(fields, "volumeId is required")
 	}
 
+	volumesClient := d.clientBlockStorage
+	var res gophercloud.ErrResult
 	if d.clientBlockStoragev2 != nil {
-		res := volumes.Delete(d.clientBlockStoragev2, volumeID)
-		if res.Err != nil {
-			return goof.WithFieldsE(fields, "error removing volume", res.Err)
-		}
+		volumesClient = d.clientBlockStoragev2
+		res = volumes.Delete(d.clientBlockStoragev2, volumeID).ErrResult
 	} else {
-		res := volumesv1.Delete(d.clientBlockStorage, volumeID)
-		if res.Err != nil {
-			return goof.WithFieldsE(fields, "error removing volume", res.Err)
-		}
+		res = volumesv1.Delete(d.clientBlockStorage, volumeID).ErrResult
+	}
+
+	if res.Err != nil {
+		return goof.WithFieldsE(fields, "error removing volume", res.Err)
+	}
+	err := waitFor404(volumesClient, volumesClient.ServiceURL("volumes", volumeID), int(d.deleteTimeout().Seconds()))
+	if err != nil {
+		return goof.WithFieldsE(fields, "error waiting for volume removal", err)
 	}
 
 	return nil
@@ -578,7 +606,7 @@ func (d *driver) VolumeAttach(
 	}
 
 	ctx.WithFields(fields).Debug("waiting for volume to attach")
-	volume, err := d.waitVolumeAttachStatus(ctx, volumeID, true, 30*time.Second)
+	volume, err := d.waitVolumeAttachStatus(ctx, volumeID, true, d.attachTimeout())
 	if err != nil {
 		return nil, "", goof.WithFieldsE(
 			fields, "error waiting for volume to attach", err)
@@ -609,7 +637,7 @@ func (d *driver) VolumeDetach(
 	}
 
 	ctx.WithFields(fields).Debug("waiting for volume to detach")
-	volume, err := d.waitVolumeAttachStatus(ctx, volumeID, false, 30*time.Second)
+	volume, err := d.waitVolumeAttachStatus(ctx, volumeID, false, d.attachTimeout())
 	if err == nil {
 		return volume, nil
 	}
@@ -621,7 +649,7 @@ func (d *driver) VolumeDetach(
 			return nil, goof.WithFieldsE(fields, "error force detaching volume", resp.Err)
 		}
 
-		volume, err = d.waitVolumeAttachStatus(ctx, volumeID, false, 30*time.Second)
+		volume, err = d.waitVolumeAttachStatus(ctx, volumeID, false, d.attachTimeout())
 		if err != nil {
 			return nil, goof.WithFieldsE(
 				fields, "error waiting for volume to force detach", err)
@@ -725,8 +753,8 @@ func (d *driver) trustID() string {
 	return d.config.GetString("openstack.trustID")
 }
 
-func (d *driver) volumeCreateTimeout() time.Duration {
-	strVal := d.config.GetString("openstack.volumeCreateTimeout")
+func (d *driver) timeout() time.Duration {
+	strVal := d.config.GetString("openstack.timeout")
 	val, err := time.ParseDuration(strVal)
 
 	if err != nil || val <= 0 {
@@ -735,12 +763,42 @@ func (d *driver) volumeCreateTimeout() time.Duration {
 	return val
 }
 
-func (d *driver) volumeSnapshotTimeout() time.Duration {
-	strVal := d.config.GetString("openstack.volumeSnapshotTimeout")
+func (d *driver) attachTimeout() time.Duration {
+	strVal := d.config.GetString("openstack.attachTimeout")
 	val, err := time.ParseDuration(strVal)
 
 	if err != nil || val <= 0 {
-		val = 10 * time.Minute
+		val = d.timeout()
+	}
+	return val
+}
+
+func (d *driver) deleteTimeout() time.Duration {
+	strVal := d.config.GetString("openstack.deleteTimeout")
+	val, err := time.ParseDuration(strVal)
+
+	if err != nil || val <= 0 {
+		val = d.timeout()
+	}
+	return val
+}
+
+func (d *driver) createTimeout() time.Duration {
+	strVal := d.config.GetString("openstack.createTimeout")
+	val, err := time.ParseDuration(strVal)
+
+	if err != nil || val <= 0 {
+		val = d.timeout()
+	}
+	return val
+}
+
+func (d *driver) snapshotTimeout() time.Duration {
+	strVal := d.config.GetString("openstack.snapshotTimeout")
+	val, err := time.ParseDuration(strVal)
+
+	if err != nil || val <= 0 {
+		val = d.timeout()
 	}
 	return val
 }
--- a/drivers/storage/openstack/storage/openstack_storage.go
+++ b/drivers/storage/openstack/storage/openstack_storage.go
@@ -1,3 +1,5 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_openstack
+
 package openstack
 
 import (
--- a/drivers/storage/openstack/storage/openstack_storage.go
+++ b/drivers/storage/openstack/storage/openstack_storage.go
@@ -89,6 +89,11 @@ func (d *driver) Init(context types.Context, config gofig.Config) error {
 	} else {
 		fields["password"] = "******"
 	}
+	if d.tokenID() == "" {
+		fields["tokenId"] = ""
+	} else {
+		fields["tokenId"] = "******"
+	}
 	fields["tenantId"] = d.tenantID()
 	fields["tenantName"] = d.tenantName()
 	fields["domainId"] = d.domainID()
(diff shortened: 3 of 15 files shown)


MatMaul:
updated, with support for a trust id as a scope. Also change from Cinder to Nova API to handle detach (inconsistency on attachments seems to appear between Nova and Cinder when using the Cinder detach API).

strigazi:
Works fine!

I created a volume using my credentials scoped with a trustID. And for our usecase in openstack/magnum, I managed to create a volume using the dedicated user associated to a cluster scoped with the trustID.

@clintonskitson @akutz 
Can you provide some feedback? I think the patches are in good shape.

akutz:
Hi @MatMaul,

Thank you so very much for this! We are prepping REX-Ray 0.5.0's RC. It will include a new libStorage release to support Amazon EFS (thanks to @mhrabovcin!) and some needed Isilon patches. I am going to target OpenStack, the inclusion of this PR, for RR 0.6.0 and the libStorage release after the one we're currently planning. I hope to have a fast turnaround. I'm tagging @cduchesne here as I'm going to ask him to test this PR.

A few things @MatMaul:

- Could you please keep the PR rebased off of master? That would help us immensely. 
- Are you interested in also providing a RackSpace driver? As I recall, the RackSpace driver is a near copy of the OpenStack driver with some minor differences. If you do not do this, I will likely handle it so we can release both the OpenStack and RackSpace drivers in the same release of libStorage and REX-Ray.

Thank you again!

akutz:
Hi @MatMaul,

One additional note - would you also mind squashing your commits into one? If you have a good reason to leave them separate, please do let me know. I'm a fan of a commit-per-PR, especially with a new feature. I like things to be atomic; it makes it easier to revert if necessary. 

This is in addition to the rebasing mentioned above. Thanks again!

MatMaul:
Hi @akutz, thanks for the feedback.

I just rebased & squashed my commits.
For the openstack vs rackspace issue I think we can write some functions wrappers here to avoid copying the whole driver and duplicate the code, I'll have a look. However if someone else can test it on rackspace afterwards it would be nice.

Last but not least : the test in error seems to be a flaky one. I already got it in failure before and it seems unrelated to my changes.

akiraw:
Hi @MatMaul,
I am currently working on and testing a Rackspace driver separate from Openstack. Your work has been very helpful and I definitely agree that they can be consolidated in the future.
As it stands now, I was hoping to get a PR in for the Rackspace driver in and then perhaps we could work together on the merging of the two.

akutz:
Hi @cduchesne,

I was curious if you've had an opportunity to test this driver?

cduchesne:
@akutz not yet - I still need to get my openstack environment up and running

MatMaul:
Any thoughts about switching to gophercloud/gophercloud instead of rackspace/gophercloud ? it is the new recommended one, and it unifies rackspace/openstack implementations. The API is almost the same. To be verified but I think there still would be some missing bits for rackspace support, for example it seems to only support volume API v1 (v2 currently used here). @akiraw could work on that for example. I would really like to avoid having 2 drivers, especially since the underlying API are very very close (the same ?).

MatMaul:
@akiraw I just had a look at your PR, your tests would be welcome too ;)

MatMaul:
I added configurable timeouts for create/snapshot actions.
You can find a gophercloud/gophercloud version here : 
https://github.com/MatMaul/libstorage/tree/change_gophercloud

akutz:
Hi @MatMaul,

@akira is an intern, and today is his last day, so no unification will be possible with him in charge. I'd like your advise here -- would you rather see us delay in merging this and the Rackspace driver for now, or would you rather you or someone else work to do the merge of the two types ahead of accepting the new drivers?

akiraw:
Either way works for me, my driver currently runs as-is with REX-Ray - volume attaches, mounts and so on. But it is redundant to have multiple drivers in the end. @akutz please go with whatever makes the most sense for the project right now.

MatMaul:
Hi all, I have a version ready with BlockStorage V1 API used as a fallback.
Regarding the tests they seem to be generic enough to be factorize, they almost only call the official libstorage API. I added them in the OpenStack driver by making the changes needed to only call the API but it should be factorized elsewhere I think.

Can someone test that ? If someone has the @akiraw environment setup somewhere it should be able to run the tests using the OpenStack driver.

here is the branch, I'll push that here if it works ;) 
https://github.com/MatMaul/libstorage/tree/change_gophercloud

MatMaul:
Ok so I pushed the changes anyway since it makes sense even without taking Rackspace merging into account. The tests are still not there, only on my change_gophercloud branch, if they pass I'll push them too.

MatMaul:
Any news regarding Rackspace testing ?

mux23:
@MatMaul I've been slaving away setting up an openstack testing environment that is more permanent. I've been able to successfully test a binary that @cduchesne sent over on an openstack environment under AWS, but work towards automating those tests basically proved that the AWS route would be prohibitively complicated and expensive.

I'll have more details soon, I believe our testing environment to be nearly ready for regular use.

MatMaul:
Cool, thanks for the heads up on the global CI.

MatMaul:
Any chances to have this merged before the next release ? I thought we were targeting next libstorage/rexray release but seems like a RC is already out :(

strigazi:
@akutz @cduchesne Any updates on reviewing?

Do you need any help on testing?

akutz:
Hi @strigazi,

I am pinging @cduchesne and @mux23 this morning. Drew was setting up the testing environment and Chris was coordinating things. I'll start reviewing the code more closely as well in parallel.

akutz:
Hi @strigazi,

According to @mux:

> I’ve been working on exactly that for a couple of weeks now. (I've been) trying to get a stable environment up and running in the lab. it’s been my primary focus - struggling with the networking currently. The shift from nova networking to neutron is a pretty big one.
> 
> I should have it stable this week

So it sounds like he should have an environment for me to futz with soon. Thank you for your patience.

cduchesne:
Hi @MatMaul - we are looking to merge this at the end of the month. @mux23 and I are in the process of testing via several methods, including Rackspace. Do you think you rebase your PR to simplify the testing?

akutz:
Hi @MatMaul,

FWIW, I'm rebasing your work onto a branch on my fork right now. I'm using `master` from the primary repo as the base instead of the current, targeted release branch. This means there will still be some additional changes required once release 0.3.3 is merged to master, but none too difficult.

akutz:
Hi @MatMaul & @cduchesne,

Please find Mat's OpenStack driver rebased onto `master` [here](https://github.com/akutz/libstorage/tree/feature/openstack-driver).

FWIW @MatMaul, you can use my work if you follow these steps:

``` bash
$ git remote add akutz https://github.com/akutz/libstorage
$ git fetch akutz
$ git checkout master
$ git reset --hard akutz/feature/openstack-driver
$ git push --force
```

The above commands will add my repo as remote for your libStorage clone, fetch my repo, checkout your copy of master (if not already checked out), reset your copy of master to reflect the work I did on my branch, and then push your copy of master back to your remote.

akutz:
Hi @MatMaul / @cduchesne,

I also just pushed [this](https://github.com/akutz/libstorage/commits/feature/openstack-driver-0.3.3) branch which rebases the already updated OpenStack driver branch onto the working release branch for libStorage 0.3.3.

akutz:
Hi @cduchesne / @mux23,

I provided @cduchesne with branches of this driver rebased onto `master` as well as the working `release/0.3.3` branch. Chri was able to successfully build the working REX-Ray 0.6.1 release branch against the latter. If we could please get an update on testing that would be great.

akutz:
Hi @MatMaul,

Could you please create a test package for the OpenStack driver? We're testing it locally and it appears to have some issues, but the problem is that the test procedure is not deterministic and repeatable. Before I run manual tests I usually like to rely on a test file for ensuring some basic level of stability. Please see the other storage drivers for examples on what I mean. Also please let me know if you need any help. Thanks!

MatMaul:
Hi @akutz,

I can build a binary and test it manually on our infra before sending it to you.
Can I have more infos on your problems and test workflow ?

akutz:
Hi @MatMaul,

Sure, if you look at any of the other driver packages you'll see they have a `test` sub-package. You should be able to copy one of those wholesale and make it work for OpenStack testing. If you have any questions, please let me know.

MatMaul:
I am talking about your real problems, not the test package. I'll write some unit tests but I am pretty confident there is no problem on this side, the code is pretty straightforward so if I expect a bug it is functional. I'll redo manually a batch of functional tests. If you have automated functional tests for rexray I would be happy to know how to launch them.

MatMaul:
My bad, they are functional tests. I'll do that.

MatMaul:
Hum I don't understand how to launch a specific set of tests, help appreciated. Doing make test only launches the common & vfs ones. go test <driver package name> or go test <driver package name/tests> don't work either.

MatMaul:
Happy new year to all ! up for my previous message :)

vladimirvivien:
@MatMaul Happy new year, apologies about the delay.  To answer your last question, you should be able to test any libstorage package with no problem.  For instance, this `make test ./drivers/storage/scaleio/...`  will test the scaleio driver.  Which test are you running that is causing the issue?

MatMaul:
Great thanks, I didn't manage to launch the tests at all. Calling make test without a path doesn't seem to run the drivers tests. I'll try that and report.

jtopjian:
Hi all,

I came across this while researching Docker volume drivers and wanted to see if I could help push it along.

tl;dr: https://github.com/jtopjian/libstorage/commit/d09a74bdf7f09d2ff0157e5907dd0374faa93e04 contains the minimum changes required to get this driver to pass the acceptance tests. I have no idea if some of the changes I made outside of the OpenStack driver are correct. I based a lot of them on the EBS driver.

The test suite is an almost direct copy of the Rackspace tests. I want to look into the suite in more detail to see if there are any enhancements that would be worthwhile. I can also do a general review of the OpenStack driver in general, _not_ because @MatMaul did things incorrectly (quite the contrary), but I'm familiar with Gophercloud so it might add some confidence to have a second set of eyes to get this merged in.

With regard to testing in general, if it wasn't for the EBS README file, I would have been absolutely lost (and was fumbling around badly for an hour or two until I found it). Here's a final set of instructions to get the test running, as well as some notes:

```shell
$ go get github.com/codedellemc/libstorage                                                               
The authenticity of host 'github.com (192.30.253.113)' can't be established.
RSA key fingerprint is SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8.
Are you sure you want to continue connecting (yes/no)? yes
# cd /root/go/src/github.com/cesanta/validate-json; git submodule update --init --recursive
Submodule 'schema/schema-tests' (git@github.com:json-schema/JSON-Schema-Test-Suite.git) registered for path 'schema/schema-tests'
Cloning into 'schema/schema-tests'...
Warning: Permanently added 'github.com,192.30.253.113' (RSA) to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
fatal: clone of 'git@github.com:json-schema/JSON-Schema-Test-Suite.git' into submodule path 'schema/schema-tests' failed
package github.com/cesanta/validate-json/schema: exit status 128
```
^ That doesn't seem right

```shell
$ cd ~/go/src/github.com/codedellemc/libstorage
$ make ./drivers/storage/openstack/tests/openstack.test
...
/bin/bash: /root/go/bin/go-bindata: No such file or directory                        
Makefile:869: recipe for target 'api/server/executors/executors_generated.go' failed 
make: *** [api/server/executors/executors_generated.go] Error 127                    
```
^ Unfortunately you can't just `go get` this because `libstorage` is using a local patched copy, so:

```shell
$ cp -a vendor/github.com/jteeuwen ~/go/src/github.com/
$ cd ~/go/src/github.com/jteeuwen/go-bindata
$ go build -o ~/go/bin/go-bindata go-bindata/*
$ cd -
$ make ./drivers/storage/openstack/tests/openstack.test
```
Now the test file should be created. If you're not running on an OpenStack instance, you can't run this test, but you can copy the binary to an OpenStack instance. That's pretty cool.

```shell
$ source ~/openrc
$ cd drivers/storage/openstack/tests
$ ./openstack.test
$ LIBSTORAGE_LOGGING_LEVEL=debug ./openstack.test  2>&1 | tee foo.log
$ LIBSTORAGE_LOGGING_LEVEL=debug ./openstack.test -test.run TestVolumeAttach 2>&1 | tee foo.log
$ LIBSTORAGE_LOGGING_LEVEL=debug ./openstack.test -test.v 2>&1 | tee foo.log
$ ./openstack.test --help
```

jtopjian:
In general, the Gophercloud implementation looks good. Snapshots aren't passing the acceptance tests and unfortunately I don't have time to dig into the reason right now. Also, Gophercloud only has support for snapshots in Block Storage v1 right now, not v2, so it might be worthwhile to throw a "not implemented" error for when the v1 client is `nil`.

@MatMaul Nice work 😄  I'd be happy to open a PR against your branch with the changes I linked to above.

MatMaul:
Hi @jtopjian, great job I was a bit out of time recently, your work really helps me :)

Feel free to open a PR against my branch, I'll merge it and see if I can fix the remaining snapshot problems.

jtopjian:
@MatMaul Can you rebase your code off of the libstorage master branch? Let me know if you need a hand with how to do that. 

I rebased https://github.com/jtopjian/libstorage/commit/d09a74bdf7f09d2ff0157e5907dd0374faa93e04 just to make sure everything was working. I can still submit a PR but I think the git history would be a lot cleaner if a rebase is done first.

MatMaul:
Rebased :)

jtopjian:
Cool! I just submitted a clean PR 😄

MatMaul:
The closing was not intended sorry. Update coming soon.

MatMaul:
For TestSnapshots : 
- SnapshotCopy is not supported (needs to be removed in the test)
- we need to wait until snapshot is actually deleted before deleting the underlying volume

1 is done, I'll fix 2 tomorrow and push.

MatMaul:
@jtopjian I think you need to sign the CLA now that a commit of you is in the PR :)

MatMaul:
The remark regarding the waiting needed in the test triggers a question : should the waiting be done inside the driver ? is the driver expected to be synchronous ?

jtopjian:
@MatMaul Signed :)

MatMaul:
Hi @akutz , did you have the time to take a look ? are you happy with the tests ?
I am waiting for an answer for the async vs sync operations at driver level but otherwise I think this is ready for inclusion.
Thanks.

Makefile
```diff
@@ -8,8 +8,10 @@ BUILD_TAGS :=   gofig \
 ifneq (true,$(TRAVIS))
 BUILD_TAGS +=   libstorage_storage_driver \
 				libstorage_storage_driver_vfs \
+				libstorage_storage_driver_openstack \
 				libstorage_storage_executor \
-				libstorage_storage_executor_vfs
+				libstorage_storage_executor_vfs \
+				libstorage_storage_executor_openstack
```
codenrhoden: Adding `openstack` here is no longer needed, now that the libStorage Makefile accepts the `DRIVERS` env var. Remove these. I would also encourage you to add a `test-openstack` and `test-openstack-clean` make target. You can find examples for other drivers at the end of this Makefile.

drivers/storage/openstack/executor/openstack_executor.go
```diff
@@ -0,0 +1,162 @@
...
+
+func (d *driver) Name() string {
+	return openstack.Name
+}
+
+// InstanceID returns the local instance ID for the test
+func InstanceID(config gofig.Config) (*types.InstanceID, error) {
+	d := newDriver()
+	d.Init(nil, config)
+	return d.InstanceID(nil, nil)
```
codenrhoden: It looks like the only reason that `newDriver()` and `d.Init()` are being called is so that you can use `d.InstanceID`. Instead, I would move the logic from `(d *driver) InstanceID` into this function, and call it from the other one.

drivers/storage/openstack/executor/openstack_executor.go
```diff
@@ -0,0 +1,162 @@
...
+func (d *driver) InstanceID(
+	ctx types.Context,
+	opts types.Store) (*types.InstanceID, error) {
+
+	uuid, err := getInstanceIDFromCloudInitFile()
+	if err != nil {
+		cloudInitFileErr := err
+		uuid, err = getInstanceIDFromMetadataServer()
+		if err != nil {
+			metadataServerErr := err
```
codenrhoden: Set this to `err.WithError("", cloudInitFileErr)`

drivers/storage/openstack/executor/openstack_executor.go
```diff
@@ -0,0 +1,162 @@
...
+			uuid, err = getInstanceIDWithDMIDecode()
+			if err != nil {
+				return nil, fmt.Errorf("%v ; %v ; %v", cloudInitFileErr, metadataServerErr, err)
```
codenrhoden: And then set this here to `err.WithError("Unable to get InstanceID from any sources", metadataServerErr)`.

I think that will work.

drivers/storage/openstack/executor/openstack_executor.go
```diff
@@ -0,0 +1,162 @@
...
+	iid := &types.InstanceID{Driver: openstack.Name, ID: strings.ToLower(uuid)}
+
+	return iid, nil
+}
+
+func getInstanceIDFromCloudInitFile() (string, error) {
+	const instanceIDFile = "/var/lib/cloud/data/instance-id"
+	idBytes, err := ioutil.ReadFile(instanceIDFile)
+	if err != nil {
+		return "", goof.WithError("error reading file "+instanceIDFile, err)
```
codenrhoden: please use `goof.WithFieldE("file", instanceIDFile, "error reading file", err)`.

drivers/storage/openstack/executor/openstack_executor.go
```diff
@@ -0,0 +1,162 @@
...
+
+	return uuid, nil
+}
+
+func getInstanceIDWithDMIDecode() (string, error) {
+	cmd := exec.Command("dmidecode", "-t", "system")
+	cmdOut, err := cmd.Output()
+
+	if err != nil {
+		return "", goof.WithError("error calling dmidecode", err)
```
codenrhoden: Unfortunately, this will just result in the `[]byte` for `stderr` to be logged, which doesn't look good in logs/errors. See [this example](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/rbd/utils/utils.go#L71-L72) for how to get a string version of stderr.

drivers/storage/openstack/executor/openstack_executor.go
```diff
@@ -0,0 +1,162 @@
...
+	ctx types.Context,
+	opts *types.LocalDevicesOpts) (*types.LocalDevices, error) {
+	devicesMap := make(map[string]string)
+
+	file := "/proc/partitions"
+	contentBytes, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil,
+			goof.WithFieldsE(
+				goof.Fields{"file": file}, "error reading file", err)
```
codenrhoden: With just one field, you can use `goof.WithFieldE()` instead. Then there is no need to use the `goof.Fields{}` struct.

drivers/storage/openstack/storage/openstack_storage.go
```diff
@@ -0,0 +1,743 @@
...
+	return nil
+}
+
+// InstanceInspect returns an instance.
+func (d *driver) InstanceInspect(
+	ctx types.Context,
+	opts types.Store) (*types.Instance, error) {
+
+	iid := context.MustInstanceID(ctx)
+	if iid.ID != "" {
```
codenrhoden: This check shouldn't really be necessary. You might have seen that the ScaleIO driver does this, but that's because for ScaleIO, the actual InstanceID can *only* be calculated on the server, not at the client (executor). Therefore the client sends an initial IID over, the server transforms it and fills in `ID`, then future calls to `InstanceInspect` see that it has already been done and it becomes a noop.

For this driver, since the executor is already sending over a fully-formed IID, this isn't necessary.

drivers/storage/openstack/storage/openstack_storage.go
```diff
@@ -0,0 +1,743 @@
...
+}
+
+func (d *driver) SnapshotRemove(
+	ctx types.Context,
+	snapshotID string,
+	opts types.Store) error {
+	resp := snapshots.Delete(d.clientBlockStorage, snapshotID)
+	if resp.Err != nil {
+		return goof.WithFieldsE(goof.Fields{
+			"snapshotId": snapshotID}, "error removing snapshot", resp.Err)
```
codenrhoden: This could be changed to `WithFieldE`

drivers/storage/openstack/storage/openstack_storage.go
```diff
@@ -0,0 +1,743 @@
...
+	options := &volumes.CreateOpts{
+		Name:             volumeName,
+		SnapshotID:       snapshotID,
+		SourceReplica:    volumeSourceID,
+	}
+
+	if opts.Type != nil {
+		options.VolumeType = *opts.Type
+	}
+	if opts.Size != nil {
```
codenrhoden, line 479: Please protect against the creation of size zero volumes in the style of https://github.com/codedellemc/libstorage/pull/495.  I know it's a pain that mass driver changes have happened while you've been working on this.

drivers/storage/openstack/storage/openstack_storage.go
```diff
@@ -0,0 +1,743 @@
...
+
+	iid := context.MustInstanceID(ctx)
+
+	fields := eff(map[string]interface{}{
+		"volumeId":   volumeID,
+		"instanceId": iid.ID,
+	})
+
+	if opts.Force {
+		if _, err := d.VolumeDetach(ctx, volumeID, &types.VolumeDetachOpts{}); err != nil {
```
codenrhoden: Please query the volume and see if it needs to be detached first.

glide.lock
```diff
@@ -117,6 +117,21 @@ imports:
...
+  - openstack/blockstorage/v1/snapshots
+  - openstack/blockstorage/v1/volumes
+  - openstack/blockstorage/v2/volumes
+  - openstack/compute/v2/extensions/volumeattach
+  - openstack/identity/v3/extensions/trusts
+  - openstack/identity/v2/tokens
+  - openstack/identity/v3/tokens
+  - openstack/utils
+  - pagination
+  - openstack/identity/v2/tenants
```
codenrhoden, line 131: modifications to `glide.lock` cannot be done this way. You will need to run `glide up` once you are happy with your `glide.yaml` so that an accurate `glide.lock` complete with updated hash is generated.

drivers/storage/openstack/tests/openstack_test.go
```diff
@@ -0,0 +1,534 @@
...
+// configuration for creating an openstack volume.
+var configYAML = []byte(`
+openstack:
+  authURL: ` + os.Getenv("OS_AUTH_URL") + `
+  username: ` + os.Getenv("OS_USERNAME") + `
+  tenantName: ` + os.Getenv("OS_TENANT_NAME") + `
+  domainName: ` + os.Getenv("OS_USER_DOMAIN_NAME") + `
+  password: ` + os.Getenv("OS_PASSWORD") + `
+  availabilityZoneName: ` + os.Getenv("OS_AVAILABILITY_ZONE") + `
+  regionName: ` + os.Getenv("OS_REGION_NAME") + `
```
codenrhoden: The calls to `os.Getenv` are not necessary. The test will already pull in any matching env vars, just the same as if you were running libStorage/REX-Ray normally. For example, the EBS driver does not pull any of those options in: https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/tests/ebs_test.go#L34-L37

But if you look at the README in the test directory, the examples show using the env vars.

codenrhoden: I would like to see a README in this test directory that has instructions on how to run the tests.  See [GCE](https://github.com/codedellemc/libstorage/tree/master/drivers/storage/gcepd/tests#unitintegration-tests) for an example that uses the Makefile targets (if you choose to add them).

MatMaul:
Thanks for the review, I'll update and answer this week.
Can I have an answer regarding the sync vs async operations in drivers so I handled that too ?
Thanks !

codenrhoden:
@MatMaul My apologies - I did see your previous question, and mean to answer it!

The drivers are expected to be synchronous -- the libStorage framework has a task system that handles asynchronous requests at a layer above the drivers. But a given driver is expected to complete a given call in a synchronous manner.

Some examples of drivers waiting for calls to finish can be seen in [EBS](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/storage/ebs_storage.go#L1143) and [GCE](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/gcepd/storage/gce_storage.go#L862). 

I would be remiss if I didn't point out that both of those implementations currently have a bug where the driver *never* returns, as documented in #480, which I will be fixing this week. If the driver enters a loop waiting on volume status to change, there does need to be a reasonable timeout associated with it, and that is currently missing.

MatMaul:
Great I was expecting it to be this way, I'll take care of keeping things synchronous.

drivers/storage/openstack/tests/openstack_test.go
MatMaul: I don't understand, you mean there is an automatic matching if I use DOMAIN_NAME for domainName for example ? if yes that's not what I want, the env vars defined here are the ones used by standard openstack RC file.

drivers/storage/openstack/executor/openstack_executor.go
MatMaul: I did something simpler.

glide.lock
MatMaul: Yes I just prefer to patch it manually when WIP to ease merging/rebasing, I'll do that.

drivers/storage/openstack/tests/openstack_test.go
codenrhoden: Yes, any configuration parameters registered for the driver also get matching ENV Vars and CLI flags automatically, thanks to Viper. This is explained pretty well in our [docs](https://github.com/codedellemc/libstorage/blob/master/.docs/user-guide/config.md#configuration-properties). Since driver configs can be placed at the root of the YAML config, that means that your `openstack.domainName` config option also has a `OPENSTACK_DOMAINNAME` env var, and a `--openstackDomainName` CLI arg.

I understand wanting to use the openstack RC file env vars. Right now you have it setup such that they will be used for testing, but they are not used at all for configuring a real instance of the driver (and I think that's good). But since you are always explicitly setting the test values to those from the RC file, it means that if we run a test that use the real, normal env vars, those will be ignored.

I wonder if it is at all feasible to only try to load the RC vars from the environment if you can verify that the parameter is not already set. In short, while I am not against using the RC vars, I still want to be able to use the driver's env vars as well. Just thinking out load on that one, not even sure if it's possible.

drivers/storage/openstack/executor/openstack_executor.go
codenrhoden: :+1: that's probably a very good thing.

drivers/storage/openstack/tests/openstack_test.go
MatMaul: Ok I'll go with the simpler : use the default mechanism and add a sh script to wrap the RC env vars.

MatMaul:
The code to get the instance id from cloud-init files is wrong : sometimes the reported id is fine, sometimes we get openstack-XXXXX instead, perhaps it is the EC2 id for the compatibility layer.

I am switching to this code (Apache licensed too):
https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/openstack/metadata.go

MatMaul:
@codenrhoden I am trying to access the OSDriver inside the executor.
I want to access the Mount primitive.
context.MustClient(ctx) is failing inside the executor.

Any other idea ?

MatMaul:
I managed to initialize properly an OSDriver looking at the client code.

drivers/storage/openstack/tests/openstack_test.go
MatMaul: I still don't understand :) how do I generate a config structure from the env vars to pass to the driver ? I can just create an empty config byte array and apitests.Run will fill it for me ?

MatMaul:
Ok so I am pushing my WIP.

What is missing : 
- make everything synchronous
- rework the config handling of the test (cf comment)
- deal with your comment on VolumeDetach, I need to check the error returned by cinder if we try to detach something not attached, perhaps we can avoid the extra checking call.

drivers/storage/openstack/storage/openstack_storage.go
MatMaul: Done.

Makefile
MatMaul: Done

drivers/storage/openstack/executor/openstack_executor.go
MatMaul: Done

drivers/storage/openstack/executor/openstack_executor.go
MatMaul: Done

drivers/storage/openstack/executor/openstack_executor.go
MatMaul: Done

drivers/storage/openstack/executor/openstack_executor.go
MatMaul: Done

drivers/storage/openstack/executor/openstack_executor.go
MatMaul: Done

drivers/storage/openstack/storage/openstack_storage.go
MatMaul: Done

drivers/storage/openstack/storage/openstack_storage.go
MatMaul: Done

glide.lock
MatMaul: Done

drivers/storage/openstack/tests/openstack_test.go
codenrhoden: I've been trying to think about this so that I don't steer you wrong. The more I've looked at the existing tests, the more I realize that most of their _built-in_ configs are totally unnecessary. It's really the first I've looked at it this closely. Aren't you the lucky one?

> I can just create an empty config byte array and apitests.Run will fill it for me ?

Yes you can. in fact it can be `nil`, as seen in the [`vbox`](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/vbox/tests/vbox_test.go#L95) driver. The tests get compiled into a binary, and when that binary executes, it will automatically look for the env vars defined by the driver's config registration.

Taking EBS as an example again, The [README](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/tests/README.md#unitintegration-tests) instructs the user to export their AWS key credentials, but those aren't in the [test config](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/tests/ebs_test.go#L34-L37) at all.

So when I really think about it, the only reason to have anything in the test file at all is if you want to explicitly give the tests a default other than what the driver already has as a default. It's also worth noting that env vars have priority over the config file, so even if something is defined in the config `[]byte` in the test file, it will be overriden if you export the appropriate variables.

Which brings me to my ultimate conclusion. If you really want to use the OpenStack RC vars by default, it actually does seem like it would be okay to do so directly in the test file. If any of those vars aren't in the environment, they just won't be set by default. However, if a test runner does export env vars that are specific to the driver, those should have precedence over what's defined there in the config string of the test file.

Does that make sense?

I know I've talked myself in a complete circle, but it was a worthwhile journey for me. :)

One other interesting point of reference is what the EFS driver does, which is to build the config [dynamically](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/efs/tests/efs_test.go#L54-L68). If I am reading through it correctly, it first looks for the standard AWS SDK env vars (much like OpenStack RC), and if those aren't present it sets some defaults. And *all* of these would be overriden if a user did something like `export EFS_ACCESSKEY...` or other valid key.

MatMaul: Ahah awesome :)
Live epiphany ;)
Make total sense, I'll keep it like that then.

MatMaul:
Ok so almost there !

@codenrhoden 
I am still getting my head around VolumeDetach.
What is the exact semantic ? should it return an error if the volume is not attached ? if no error, should I still return the unattached volume descriptor as a result ?
The comment on the interface doesn't say, I don't know if there is any other ref for the API.

codenrhoden:
Hi @MatMaul,

Almost all drivers return an error if the volume is not attached. See [EBS](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/storage/ebs_storage.go#L675) and [RBD](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/rbd/storage/rbd_storage.go#L323) for examples.

You are right that there needs to be a better reference for the API, a "best practice" for a driver. That's ultimately what #384 will turn into, though It's not currently in that state.

At an API level for the libStorage server, meaning the RESTful interface, there are [apiary docs](http://docs.libstorage.apiary.io/#), but I don't think that's what you are looking for. We recognize the shortcoming here, and and working on making improvements.

MatMaul:
Just some comments in the interface for some edge cases like this one would be enough ;) 
Final review coming soon.

MatMaul:
Last question (I swear :) ):
If VolumeAttach is called and the volume is already attached to the requested server should I error out or just do nothing and return ok ?

MatMaul:
Looking at EBS code it looks like it would error out if !forced and it is attached, even if attached on the specified server. I am not sure I agree with that :)

codenrhoden:
> Looking at EBS code it looks like it would error out if !forced and it is attached, even if attached on the specified server. I am not sure I agree with that :)

That's a fair point. You are interpreting what you see correctly, though. The reason it is this way is that it is assumed that we won't get an attach for a volume to a host where it is already attached. I think a real argument can be bade on whether that assumption is valid. The reason it works okay today is because of REX-Ray. Before an attach occurs, REX-Ray will [request a list of volumes](https://github.com/codedellemc/rexray/blob/master/cli/cli/cmds_10_volume.go#L288-L295) that are available to be attached (meaning, volumes that are *available*). If a volume is already attached, it is not available. This behavior can be modified via flags, such as `-i`, whereby `rexray volume attach -i volname1` called repeatedly on the same host will succeed each time.

That idempotent flag went in with https://github.com/codedellemc/rexray/pull/658

Hope that alleviates some fear.

But your point is still taken. REX-Ray is not intended to be the only client of libStorage. Safety measures that REX-Ray takes should not validate assumptions in libStorage. We have tried to be conscious of this, but I'm sure there are some skeletons in the closet. :)

MatMaul:
I understand what you say, but I still don't get if the EBS behavior is expected or kind of a bug (that we don't really care because not visible in rexray). Basically I still don't know if it should fail in the same way in the openstack driver. I see some points from @akutz in other related bugs regarding possible extra unneeded calls, but it is not the case here (in EBS or in OpenStack), we do the check call anyway. Should we fix the EBS behavior ?

MatMaul:
My bad it is not true we don't check in the OpenStack driver, and calling 2 times attach currently fails as with EBS, as Cinder gives me back a 400 error if I try that. So let's keep it like that, at least it is coherent.

Now back to the VolumeDetach call in attach made when force is true : you asked me to add a check, what about just ignore any error coming back from VolumeDetach ? If something goes wrong on detach you will see it in the attach call afterwards anyway, and it avoids an extra call.

codenrhoden:
> Now back to the VolumeDetach call in attach made when force is true : you asked me to add a check, what about just ignore any error coming back from VolumeDetach ? If something goes wrong on detach you will see it in the attach call afterwards anyway, and it avoids an extra call.

Yeah I see your point. But what if the volume needs to be detached, and that detach actually fails? If we ignore the failure, the code will proceed with attaching the volume to another host when we don't want it to. I'm not sure what Cinder will do with such a request, as this is storage provider dependent, and Cinder is another layer on top of that. Some products will allow you to attach block devices in multiple places and happily let you corrupt your filesystem if you choose to do so.

I think I'd rather see the check, or have `VolumeDetach` do the check and return an appropriate error that `VolumeAttach` can check for, correctly ignoring a `volume not attached` error. Either way.

MatMaul:
Agreed, with cinder being vendor/technology-neutral the exact behavior of the API can be a bit tricky... Final code coming with a check added.

drivers/storage/openstack/storage/openstack_storage.go
codenrhoden: Hi @MatMaul. One nit here. This code will have the behavior of always making sure that the minimum size is used, even if the user hasn't requested a size (as indicated by `opts.Size` being `nil`. What we're really looking for is: If no size is given, default to the minimum and proceed. If a size *is* given, make sure it meets the minimum, and if not return an error. See [GCE](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/gcepd/storage/gce_storage.go#L379-L389) for an example of this.

MatMaul:
@akutz I think you broke some stuffs with your last changes on paths and tests :) 
The code of TestServices seems the same as in the EBS driver.
```
=== RUN   TestServices
--- FAIL: TestServices (0.00s)
panic: interface conversion: interface {} is nil, not *types.PathConfig [recovered]
	panic: interface conversion: interface {} is nil, not *types.PathConfig

goroutine 159 [running]:
testing.tRunner.func1(0xc4201af1e0)
	/usr/local/go/src/testing/testing.go:622 +0x29d
panic(0x9cd320, 0xc4234ed340)
	/usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/codedellemc/libstorage/api/context.MustPathConfig(0xd5ea60, 0xc4234e6c60, 0xc4234e6c60)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/context/context.go:325 +0xb5
github.com/codedellemc/libstorage/api/utils.GetTempSockFile(0xd65200, 0xc4234e6c60, 0xc420633a40, 0x1)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/utils/utils.go:35 +0x60
github.com/codedellemc/libstorage/api/tests.initTestConfigs(0xd65200, 0xc4234e6c60, 0xc423501bc0)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/tests/tests.go:580 +0x1a2
github.com/codedellemc/libstorage/api/tests.getTestConfigs(0xd65200, 0xc4234e6c60, 0xc4201af1e0, 0xa6e12d, 0x9, 0xd63100, 0xc420454dc0, 0xc420454dc0, 0x0, 0x0, ...)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/tests/tests.go:513 +0x472
github.com/codedellemc/libstorage/api/tests.(*testHarness).run(0xc420454da0, 0x0, 0x0, 0xc4201af1e0, 0x1, 0x0, 0xa6e12d, 0x9, 0xc42006ef70, 0xc3, ...)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/tests/tests.go:401 +0x1ac
github.com/codedellemc/libstorage/api/tests.run(0x0, 0x0, 0xc4201af1e0, 0x1, 0x0, 0xa6e12d, 0x9, 0xc42006ef70, 0xc3, 0xd0, ...)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/tests/tests.go:314 +0x155
github.com/codedellemc/libstorage/api/tests.Run(0xc4201af1e0, 0xa6e12d, 0x9, 0xc42006ef70, 0xc3, 0xd0, 0xc42000e0e8, 0x1, 0x1)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/api/tests/tests.go:125 +0xc5
github.com/codedellemc/libstorage/drivers/storage/openstack/tests.TestServices(0xc4201af1e0)
	/home/mvelten/git/rexray/src/github.com/codedellemc/libstorage/drivers/storage/openstack/tests/openstack_test.go:137 +0xc0
testing.tRunner(0xc4201af1e0, 0xa8fe70)
	/usr/local/go/src/testing/testing.go:657 +0x96
created by testing.(*T).Run
	/usr/local/go/src/testing/testing.go:697 +0x2ca
```

akutz:
I did. Travis pointed it out. But that's why I merged a retro fix for that and made my updates with new func sigs and reverted the original sigs back to the way they were. Sorry for that.

MatMaul:
My point is that it was not compiling at all before, now it fails at runtime with current master.

codenrhoden:
@MatMaul PR #517 should take care of the issue you noticed first. Thanks!

MatMaul:
And here is the log : 
https://gist.github.com/MatMaul/99659f54b51777340c6db7c514d1117a

I filtered out the numerous "removing local device w/ invalid volume id" warning.

codenrhoden:
HI @MatMaul,

Thanks for the gist. I'm happy to see the tests pass for you. :)

Do you expect that the driver and tests should work with a DevStack deployment? I've spun up the simplest version of DevStack on a single node, and can't make any progress. I've tried executing the test binary, and using `rexray` that I compiled with your driver in it. 

For the test binary, I know you set things up such that it should read the credentials and config from the `OS_` vars that are exported by `openrc`, but that doesn't seem to work for me.  Here's a gist of that: https://gist.github.com/codenrhoden/b3fd850589e348db54c90f8290be92dd

No idea where the `<` character is coming from there.

Alternatively, if I use `rexray`, it also fails authentication, but with different logs. Here is the gist for that, including my config (I created by config based on your sample docs): https://gist.github.com/codenrhoden/e136b19625063338adc44ac5b7c51d4c

I think it's worth noting that if I do `openstack --debug volume list`, and see what settings it is using, it ultimately spits out `Using parameters {'username': 'demo', 'project_name': 'demo', 'user_domain_id': 'default', 'auth_url': 'http://10.138.0.2/identity_admin', 'password': '***', 'project_domain_id': 'default'}`

I think this seems to match my config just fine. So I'm really not sure where the auth issues are coming from.....

codenrhoden:
Two things not related to testing...

Before merging, we'll want the commits squashed. I know that there are multiple people involved in this patch, so if you want to squash into a couple different patchsets by different people, that's acceptable.

The other is something I missed during code review. The executor has not implemented the `Supported` method. Is this on purpose, or just an oversight? It stood out to me in your test output. We generally use this method to contain any checks we can perform locally to see if we are on a host where the storage functions are likely to succeed. For cloud providers, this is usually some test of "are we on the cloud provider's infrastructure"? This can usually be done by seeing if we can hit said provider's prescribed metadata server. For other drivers, it can be checking that certain required binaries are available, kernel modules loaded, etc.

MatMaul:
For devstack testing please use the stable/ocata branch, otherwise you get master and potential bugs with it.
The error is really weird so I'll not try to understand for now :)

MatMaul:
The auth url is the admin endpoint it's weird, try to source like that : 
```source devstack/openrc demo demo```
Notice the repeated demo ;)

**Comment by MatMaul** [comment]
I'll squash as much as I can but I don't want to loose ownership and some commits make sense on their own (fsType one for example), and I don't want to spend 2h reordering and rebasing so you will still have 5-6 commits.

**Comment by strigazi** [comment]
@codenrhoden Try like so:

```
$ cat /etc/rexray/config.yml 
rexray:
  logLevel: debug
libstorage:
  service: openstack
openstack:
  authURL: http://10.138.0.2/identity/v3
  userName: demo
  password: rexray
  tenantName: demo
  domainName: default
  availabilityZoneName: nova
  regionName: RegionOne
```

**Comment by codenrhoden** [comment]
I have a bit more to do, but I just wanted to say thanks @MatMaul and @strigazi, as I am starting to have success. :) I'm out of practice w/ OpenStack, so there has been a learning curve, but I got the first test to pass today, and will try running the full suite here very soon.

**Comment by codenrhoden** [comment]
Hi @MatMaul,

So this is looking pretty good from my end. I can run the test suite, and I am only getting one failure. It's a really strange failure, and not one you encountered. So I'd like to see if you run things today, whether you see it or not. If I build from your current tip (https://github.com/MatMaul/libstorage/commit/d24ccafdfc7302320c3601e156eb316f18e0e08d), and run `./openstack.test -test.run TestSnapshots`, I get the following error:

```sh
        Error Trace:    openstack_test.go:409
	            	openstack_test.go:377
	            	tests.go:434
	            	asm_amd64.s:2197
	Error:      	"map[]" does not contain "openstack"
```

This is coming from https://github.com/MatMaul/libstorage/blob/master/drivers/storage/openstack/tests/openstack_test.go#L409, and is definitely something that should be passing. This actually shouldn't be driver specific, and I will try to recreate with a different driver tomorrow. Just curious if it still passes for you.

What makes it weird to me is that if I build a full `REX-Ray`, and then hit the API manually, the response is definitely correct:

```sh
$ curl localhost:7979/snapshots
{
  "openstack": {
    "36fa1b4a-3c12-4c76-a1b1-cc5a2780d40e": {
      "name": "libstorage test snapshot",
      "id": "36fa1b4a-3c12-4c76-a1b1-cc5a2780d40e",
      "startTime": 1493262728,
      "status": "available",
      "volumeID": "c15ab594-b579-4e40-a3e0-acfc8e643ae8",
      "volumeSize": 1
    },
    "e4a5a9f6-c320-476a-bef5-8ea5c2aa1d74": {
      "name": "libstorage test snapshot",
      "id": "e4a5a9f6-c320-476a-bef5-8ea5c2aa1d74",
      "startTime": 1493249672,
      "status": "available",
      "volumeID": "07bb80d5-88c2-477e-9acd-b9d6bc953ecc",
      "volumeSize": 1
    }
  }
}
```

So I really don't understand why this is failing. Otherwise, I think we are looking good. Nice work!

codenrhoden:
@akutz Except for the one test failure I'm getting in my previous message, I think this is ready. It does include a change that allows the local OS driver to handle arbitrary filesystem types when mounting, and that's separated into it's own commit, https://github.com/codedellemc/libstorage/pull/243/commits/c6d266463eb16bad3ce67dba83698b81fb6669f5.

This is to support mounting OpenStack's "config drive", which is usually an ISO 9660 filesystem.

As for the single test failure I am getting, I am still trying to track it down. When the test harness calls `client.API().Snapshots()`, it's getting an empty map back. This is bizarre because if I pull this driver into REX-Ray and launch and hit the same API with curl, it comes back correct. More interestingly, `Volumes()` doesn't have this problem, and the contributors have run the tests successfully (see attached gist).

What's even weirder is that I quickly put together Snapshot support for the gcepd driver (just returning an empty slice instead of `notSupported`), and duplicated this test, and it passed! I'll keep digging into this one part of things.

akutz:
Hi @MatMaul,

Looks great! There are some items need to be addressed:

* The commits on this PR need to be squashed. I know you've told @codenrhoden you have some concerns about that, and that's fair. However, PRs are atomic affairs; especially with regards to a new a driver. Squash the commits and retain authorship the way you wish so that the right people are mentioned as contributors. But however you slice this pie, the commits must be squashed before we merge this. It's not about being able to add a feature; it's about being able to remove it cleanly should it ever be necessary.

* I'm not comfortable with the executor using the OS driver to mount a remote disk to discover an instance ID. @codenrhoden mentioned there is a metadata server available, but it's not used as a first-choice solution. It should be. There is no other driver that mutates the system to discover the instance ID. It's supposed to be a harmless operation. Please either make a case for the current solution or please switch to the metadata URL.

* Why is this called an OpenStack driver if the underlying SDK is compatible with OpenStack and Rackspace? Shouldn't all the paths and driver name be renamed to Cinder?

This is looking good, and we appreciate your work on this. Thanks again!

akutz:
This PR has been replaced by #536.

MatMaul:
@akutz I didn't see your last comment, here is some answers :
- It would nice to keep a separated commit for the tests with Joe as the author, squashing is nice but I don't like to completely lose the authorship in the git metadata. Also squashing my generic commit regarding fsType in the middle of the driver inclusion looks bad to me, but it is your repo not mine :) I can open a different PR for that if you prefer. Also we lose all the comments I made on the improvements. I'll push here something I think is better.
- for metadata I am fine to switch to server first, I don't think I have any rational to do the config driver first. I'll change that. Perhaps it was because by default OpenStack uses config drive.
- Cinder vs OpenStack I don't really care. I don't think Cinder is less OpenStack specific but let's keep it since it is coherent with the naming of the Amazon driver, and on the principle we could have a Manila driver (shared FS OpenStack abstraction). For Rackspace compatibility however it is untested, perhaps some bits are missing. If you have an account to test please do, then if ok we can remove the Rackspace specific one.

@codenrhoden for the ```Supported``` method I would need to do the same work as in InstanceID so if errors coming from InstanceID are correctly handled I don't think it is useful. Let me know.

jtopjian:
I very much appreciate @MatMaul's pushback to make sure my authorship is maintained in this feature. However, if it would help move this feature along, let this be explicit permission to squash as needed. 

I recognize and appreciate that credit is given in the commit message and I respect individual styles and policies of each open source project. Perhaps an idea in the future to help get around this issue, maybe tag commit messages with something like:

```
driver/cinder: something something
driver/cinder: another piece of something
```

This should make it easy to identify an overarching feature across multiple commits.

MatMaul:
Thanks for your input. I basically redid the git history from the files, and rename everything to Cinder at the same time. I put a separate commit with you as author for the tests implementation.
If @akutz still prefer one commit I'll squash them like he did.
Push coming soon.

akutz:
Hi @MatMaul & @jtopjian,

I am empathetic. Ask anyone on the project; I am not one to withhold credit and am usually the biggest cheerleader when it comes to espousing the contributions of others. Heck, read the [release note](https://github.com/codedellemc/libstorage/releases) "Thank You" sections :)

The fact of the matter is that we are planning a release for tomorrow and are behind, in part because we were waiting on feedback on this PR. @codenrhoden and I made the decision to pull the PR, squash it, and take it across the finish line in order to ensure it's included in the release.

With regards to credit, have y'all looked at the actual commit I created when I squashed this PR? The PR itself:

image

The commit message:

image

I was careful to ensure that @MatMaul is still the author and made sure to include @jtopjian in the message as a Git commit cannot have more than one author. If this is not satisfactory then I can create a string with you as an author pair and commit with the author set to that.

jtopjian:
@akutz Yep, I do see that credit is in the commit message -- thanks! It's quite possible we were both typing a message at the same time, so just a heads up that I have given permission above to squash :)

akutz:
Hi @MatMaul & @jtopjian,

I do believe in a single, atomic commit for pulling in a PR, and authorship doesn't trump that. The reason isn't for cleanliness in the merge but rather the neatness of a possible revert if there is a critical issues discovered post-inclusion.

@MatMaul, I just reached out to @codenrhoden to solicit his opinion. We're already behind on the RC, and he stayed up late working on a second patch for this PR. The second PR is ready to go, and I was about to squash the Cinder rename into your commit and prepare a merge. We need to get an RC out the door this morning, so I cannot guarantee we'll be using *this* PR to do that. 

I really apologize for the situation, and perhaps we should have followed up with you after my message last week, but at this point we're staring down the barrel of the release schedule.

akutz:
Hi @MatMaul,

I am not guaranteeing this PR will be used in lieu of the one I prepared yesterday. But in case it is, I ask that you prepare this PR as two commits. One of yours, one of @jtopjian's. The trick is you also need to include the work @codenrhoden and I did in 7fefc48589b9344fd565078cde82a91443426c21. But that needs to be a part of one of your two commits, so you need to figure out the ordering since if you apply it to your commit it may mess up the second commit, which I would assume is @jtopjian's. We were about to squash 7fefc48589b9344fd565078cde82a91443426c21 into 016d23b12cee1ac1357ef88a9b59bc6181b0e421 and merge, but I reached out to @codenrhoden for his opinion. Once we have a discussion I'll let you know. Having a PR prepared and ready would be helpful.

MatMaul:
https://github.com/MatMaul/libstorage/commits/master

MatMaul:
It didn't update the PR since it is closed I guess. Let me know if that suits your needs. I haven't test rexray itself but the tests are fine. I switched cloud config drive and metadata server. My previous work was also missing a commit to address the remark regarding the min size. Let me know, I don't move until it is merged ^^

akutz:
Hi @MatMaul,

I will speak with @codenrhoden. It appears GH won't allow me to reopen this commit since you force pushed. I force push all the time, so this is annoying. I guess it only affects PRs once they're closed.

If we take this I will simply update my PR by resetting it to your HEAD commit. I've had your repo as a configured remote for a few months now :)

MatMaul:
Looking at your refactor I think I am just missing the part where you add string consts for the driver config. It is a copy paste for me so let me know if you prefer that I'll rebase.

akutz:
Hi @MatMaul,

There are a few issues, and I ran into them as well when recreating your PR, where I fixed them:

* The documentation should be ordered alphabetically with other providers. See my PR's [example](https://github.com/akutz/libstorage/blob/feature/cinder/.docs/user-guide/storage-providers.md#cinder).
* The documentation's YAML config example should match mine as it includes updates by @codenrhoden from #531.
* The storage driver package should be `storage`, not `cinder`. Fix that and any places that imports it. For example, your `imports_remote.go` imports `github.com/codedellemc/libstorage/drivers/storage/cinder/storage`, but that doesn't exist.
* Your `glide.lock` and `glide.yaml` files are FUBAR. Please add my remote, fetch it, and `git checkout 7fefc48589b9344fd565078cde82a91443426c21 glide.yaml glide.lock` to make them match mine.

akutz:
Hi @MatMaul,

Honestly, you know what? The fastest path to this will be if you or I take my PR and split it into the same three commits you have in this branch. One for your FS update, one for the driver, and one for @jtopjian's commits. If you think you can do that using a squashed version of my PR (it has a commit for your work and then the rename) then let me know. I say this because @codenrhoden had declared that PR ready for merge late last night.

MatMaul:
Ok I'll do that, give me 15mn.

akutz:
Hi @MatMaul,

Sure. I am trying to get one other feature over the finish line as well. That's why I groaned when I saw you finally respond early this morning. I didn't want to ignore you, but I also know my time is precious at the moment :)

MatMaul:
Done. Tests ok. Code changed for size handling and metadata server vs cloud drive.

akutz:
Hi @MatMaul,

Wow, great work!

So, [this](https://github.com/MatMaul/libstorage/commits/master) branch then? 

Also, would you update the commit message subject for 296a4dd26e5b259367669b759b7a3dec41641cb8 to be `Cinder storage driver`. I ask because it's been a consistent pattern used for the addition of new drivers. The message's body is fine. The subject just makes searching for the initial commit of storage drivers easier :)

akutz:
Hi @codenrhoden,

As far as I can tell, @MatMaul's updated branch looks pretty good. Here's a diff between mine and his:

```bash
$ git diff origin/feature/cinder matmaul/master
```

```diff
b/drivers/storage/cinder/executor/cinder_executor.go
index 58b64a3..b9a2848 100644
--- a/drivers/storage/cinder/executor/cinder_executor.go
+++ b/drivers/storage/cinder/executor/cinder_executor.go
@@ -58,12 +58,12 @@ func (d *driver) Name() string {
 
 func (d *driver) InstanceID(ctx types.Context, opts types.Store) (*types.InstanceID, error) {
 	fields := map[string]interface{}{}
-	uuid, err := getInstanceIDFromConfigDrive(ctx, d)
+	uuid, err := getInstanceIDFromMetadataServer()
 	if err != nil {
-		fields["configDrive"] = err
-		uuid, err = getInstanceIDFromMetadataServer()
+		fields["metadataServer"] = err
+		uuid, err = getInstanceIDFromConfigDrive(ctx, d)
 		if err != nil {
-			fields["metadataServer"] = err
+			fields["configDrive"] = err
 			uuid, err = getInstanceIDWithDMIDecode()
 			if err != nil {
 				fields["dmidecode"] = err
diff --git a/drivers/storage/cinder/storage/cinder_storage.go b/drivers/storage/cinder/storage/cinder_storage.go
index 91916f5..f8d95cd 100644
--- a/drivers/storage/cinder/storage/cinder_storage.go
+++ b/drivers/storage/cinder/storage/cinder_storage.go
@@ -473,31 +473,37 @@ func (d *driver) createVolume(
 		SourceReplica: volumeSourceID,
 	}
 
+	fields := eff(map[string]interface{}{
+		"volumeName":     volumeName,
+		"snapshotId":     snapshotID,
+		"volumeSourceId": volumeSourceID,
+	})
+
 	if opts.Type != nil {
 		options.VolumeType = *opts.Type
 	}
-	if opts.Size != nil {
-		options.Size = int(*opts.Size)
-		if options.Size < minSizeGiB {
-			options.Size = minSizeGiB
-		}
-	}
+	fields["volumeType"] = options.VolumeType
+
 	if opts.AvailabilityZone != nil {
 		options.AvailabilityZone = *opts.AvailabilityZone
 	}
-
 	if options.AvailabilityZone == "" {
 		options.AvailabilityZone = d.availabilityZone
 	}
+	fields["availabilityZone"] = options.AvailabilityZone
 
-	fields := eff(map[string]interface{}{
-		"volumeName":       volumeName,
-		"snapshotId":       snapshotID,
-		"volumeSourceId":   volumeSourceID,
-		"volumeType":       options.VolumeType,
-		"size":             options.Size,
-		"availabilityZone": options.AvailabilityZone,
-	})
+	if opts.Size == nil {
+		size := int64(minSizeGiB)
+		opts.Size = &size
+	}
+
+	if *opts.Size < minSizeGiB {
+		fields["minSize"] = minSizeGiB
+		return nil, goof.WithFields(fields, "volume size too small")
+	}
+
+	options.Size = int(*opts.Size)
+	fields["size"] = *opts.Size
 
 	if d.clientBlockStoragev2 != nil {
 		volume, err := volumes.Create(d.clientBlockStoragev2, options).Extract()
```

If you want to verify this @codenrhoden, I think we can take @MatMaul's work. It's up to you, you did the testing. @MatMaul, does the above diff look right to you?

MatMaul:
Ok. I am also getting the same error as @codenrhoden with snapshots now...
I think we are missing something extremely stupid here, we'll see later.

I can confirm the diff is ok.

akutz:
Hi @MatMaul,

Seeing as how REX-Ray has snapshots disabled from a CLI perspective, I think we're okay for now. It's something we want to fix before we introduce them widely again however.

codenrhoden:
The above diff looks good to me. They are both changes we requested.

And yeah, I have no idea what is going on with that snapshot testing error. Still hit it last night. I don't think it's a show-stopper, though. Hitting the API via curl produces correct results.

akutz:
Hi @MatMaul, @jtopjian, and @codenrhoden,

Please see #536's [commits](https://github.com/codedellemc/libstorage/pull/536/commits) list :)

MatMaul:
LGTM. Merge incoming ! It's beer o'clock !

akutz:
Hi @MatMaul,

Indeed. As soon as the associated Travis-CI build completes, I will be encouraging @codenrhoden to push that big 'ol merge button :)

akutz:
Hi All,

Ironically Travis-CI only shows the author of the most recent commit :)

image

akutz:
Huzzah!

image

akutz:
Hi @MatMaul & @jtopjian,

Thank you both again. I'll have to come up with something extra fun for the "Thank You" section for you two.

FYI @MatMaul, I am going to refactor your OS driver work at some point, likely for the next MINOR release. You may not have realized it, but the LS Client inserts the OS, Integration, and Storage drivers into the context so any of them can pull one of the others out if they know their key. That's why I use proxies for them, so I can make sure each of the three has access to each of the other three.

Since we must support Config drives (ugh), I'm going to instantiate the OS driver as part of the LSX start up and simply insert it into the context so your executor can pull it out. I'd rather do this since your code is importing those packages anyway at this point. That way it's centralized.

MatMaul:
Yep I first tried to use the OS driver in the executor the same way as in the client but failed because not initialized hence registered so I fallback to this. Agreed that it would be better to do it globally.

MatMaul:
I now have a contributor badge, I am happy !

akutz:
So does @jtopjian :)

MatMaul:
My split was not in vain !

akutz:
Hi @MatMaul and @jtopjian,

I hope you enjoy the [Thank Yous](https://github.com/codedellemc/libstorage/releases/tag/v0.6.0) I took to the time to craft about you both :)

jtopjian:
@akutz Brilliant 😄

MatMaul:
ahahahah I didn't expect something like that at all !
Thanks it makes my day :)

Issue #245, opened 2016-08-19
Provide option to modify root user mapping with Isilon storage driver


cduchesne:
Currently, the Isilon storage driver does not allow modification of root user mapping options. This can prevent docker from having correct read/write permissions when accessing a mounted volume, as the root user privileges will be squashed.

The primary changes will be made in the emccode/goisilon project and then updated in emccode/libstorage.

Dependent Issues:
emccode/goisilon#18

cduchesne:
The configuration changes will make it so that non-root users map to root. This will allow docker containers that run processes as a non-root user to be able to write to the appropriate nfs mount.

cduchesne:
This feature has been added through PRs #251, #252 and #253

Issue #244, opened 2016-08-19
Provide option to configure Root Clients with Isilon storage driver


cduchesne:
Currently, the Isilon storage driver creates all exports with client IPs specified in the Clients parameter. This prevents root-level access to mounted volumes, and will block docker using an Isilon volume. A change is needed to allow configuring a client IP as part of the Root Clients parameter.

The primary changes will be made in the emccode/goisilon project.

Issues:
emccode/goisilon#16

cduchesne:
Through testing, this is no longer necessary to do. We were able to get docker containers with both root and non-root users to make changes to the nfs mounts via mapping non-root users to root and mounting as standard clients.

PR #250, opened 2016-08-26
Support for GoIsilon v1.1.0

akutz:
This patch adds support for the GoIsilon v1.1.0 release, enabling support for custom volume paths and enhanced NFS export security settings.

Note: This update does break backwards compatibility for the Isilon driver's `volumePath` configuration key. This key's value used to be appended to the base path `/ifs/volumes`, but that's no longer the case. The value of the `volumePath` key is now considered to be absolute. If it is not provided or empty, the value `/ifs/volumes` will be used. But if a value is specified for the key, that is the value used for the volume path, sans any prefix.

Related to #244, #245, #246.


akutz:
Replaced with #252

PR #287, opened 2016-09-30
EBS Session Caching

akutz:
This patch introduces EBS session caching. Now that EBS sessions are scoped to HTTP requests and shuttled around with the Context object, it becomes necessary to optimize their usage. Session objects are now cached using a key that is the MD5 checksum of the concatenation of the following string values that are identifying characteristics of an EBS session: 

* region
* endpoint
* access key

Any unique combination of the above values (including if one is absent) will produce a unique session object that is then cached for future use, available to requests with the same identifying characteristics for establishing an EBS session.

### DIFF
--- a/drivers/storage/ebs/storage/ebs_storage.go
+++ b/drivers/storage/ebs/storage/ebs_storage.go
@@ -1,7 +1,9 @@
 package storage
 
 import (
+	"crypto/md5"
 	"fmt"
+	"hash"
 	"strings"
 	"sync"
 	"time"
@@ -39,6 +41,7 @@ type driver struct {
 	region     *string
 	endpoint   *string
 	maxRetries *int
+	accessKey  string
 }
 
 func init() {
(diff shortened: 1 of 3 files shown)


drivers/storage/ebs/storage/ebs_storage.go
```diff
@@ -77,43 +81,70 @@ func (d *driver) Init(context types.Context, config gofig.Config) error {
...
-		sess     = session.New()
+		ckey     string
+		hkey     = md5.New()
+		akey     = d.accessKey
+		region   = d.mustRegion(ctx)
 	)
 
-	if region = d.mustRegion(ctx); region != nil {
-		fields["region"] = *region
+	if region != nil {
```
codenrhoden: Is this possible after the use of `d.mustRegion()` above? That function makes sure region is set in the IID fields, and that is is not an empty string (`""`), but can it still be `nil`?

akutz:
Yes, because if it's not in the IID or if it is and empty it returns d.region, which can itself be nil since it's a pointer to a string.

-- 
-a

PR #296, opened 2016-10-12
Fix for Executor when Not Implemented

akutz:
This patch should provide a means for a platform's executor code to indicate that a function is not implemented on a given system. This way the libStorage client may check for the "not implemented" error upon initialization and not fail to start if one of storage platforms available on a remote libStorage service is invalid for the current client.

For example, if a remote libStorage server is configured with three services -- two for EBS and one for ScaleIO -- and a client that *only* supports EBS accesses the libStorage server, the client would fail upon initialization. This is because the client scans the server to determine what services (and their drivers) are configured and upon initialization attempts the discovery of the remotely configured platform's instance information locally through the use of the storage executor binary. Well, just because the remote libStorage server has a ScaleIO service, not all clients are required to support that. Yet all clients would see that service and attempt to discover their ScaleIO instance information. 

Now executor implementations have the ability to determine if they are supported on a given client and return "not implemented" (`types.ErrNotImplemented`) if they are not. This error will not cause the client to fail, but rather simply emit a message to the log indicating that it occurred.

### DIFF
--- a/api/types/types_drivers_executor.go
+++ b/api/types/types_drivers_executor.go
@@ -11,6 +11,15 @@ import (
 type DeviceScanType int
 
 const (
+	// LSXExitCodeNotImplemented is the exit code the executor binary uses to
+	// indicate a function is not implemented for a given storage driver on the
+	// current system.
+	LSXExitCodeNotImplemented = 2
+
+	// LSXExitCodeTimedOut is the exit code the executor binary uses to indicate
+	// a function timed out.
+	LSXExitCodeTimedOut = 255
+
 	// LSXCmdInstanceID is the command to execute to get the instance ID.
 	LSXCmdInstanceID = "instanceID"
 
@@ -24,6 +33,10 @@ const (
 	// LSXCmdWaitForDevice is the command to execute to wait until a device,
 	// identified by volume ID, is presented to the system.
 	LSXCmdWaitForDevice = "wait"
+
+	// LSXCmdSupported is the command to execute to find out if an executor
+	// is valid for a given platform on the current host.
+	LSXCmdSupported = "supported"
 )
 
 const (
@@ -122,6 +135,21 @@ type StorageExecutorFunctions interface {
 		opts *LocalDevicesOpts) (*LocalDevices, error)
 }
 
+// StorageExecutorWithSupported is an interface that executor implementations
+// may use by defining the function "Supported(Context, Store) (bool, error)".
+// This function indicates whether a storage platform is valid when executing
+// the executor binary on a given client.
+type StorageExecutorWithSupported interface {
+	StorageExecutorFunctions
+
+	// Supported returns a flag indicating whether or not the platform
+	// implementing the executor is valid for the host on which the executor
+	// resides.
+	Supported(
+		ctx Context,
+		opts Store) (bool, error)
+}
+
 // ProvidesStorageExecutorCLI is a type that provides the StorageExecutorCLI.
 type ProvidesStorageExecutorCLI interface {
 	// XCLI returns the StorageExecutorCLI.
@@ -131,7 +159,7 @@ type ProvidesStorageExecutorCLI interface {
 // StorageExecutorCLI provides a way to interact with the CLI tool built with
 // the driver implementations of the StorageExecutor interface.
 type StorageExecutorCLI interface {
-	StorageExecutorFunctions
+	StorageExecutorWithSupported
 
 	// WaitForDevice blocks until the provided attach token appears in the
 	// map returned from LocalDevices or until the timeout expires, whichever
--- a/api/types/types_errors.go
+++ b/api/types/types_errors.go
@@ -8,6 +8,9 @@ import (
 // a function is not implemented.
 var ErrNotImplemented = goof.New("not implemented")
 
+// ErrTimedOut is the error that is used to indicate an operation timed out.
+var ErrTimedOut = goof.New("timed out")
+
 // ErrUnsupportedForClientType is the error that occurs when an operation is
 // invoked that is unsupported for the current client type.
 type ErrUnsupportedForClientType struct{ goof.Goof }
--- a/cli/lsx/lsx.go
+++ b/cli/lsx/lsx.go
@@ -23,7 +23,7 @@ import (
 
 var (
 	cmdRx = regexp.MustCompile(
-		`(?i)^instanceid|nextdevice|localdevices|wait$`)
+		`(?i)^supported|instanceid|nextdevice|localdevices|wait$`)
 )
 
 // Run runs the executor CLI.
@@ -60,7 +60,6 @@ func Run() {
 	if cmd == "" {
 		printUsageAndExit()
 	}
-	cmd = strings.ToLower(cmd)
 	store := utils.NewStore()
 
 	var (
@@ -69,7 +68,19 @@ func Run() {
 		exitCode int
 	)
 
-	if cmd == "instanceid" {
+	if strings.EqualFold(cmd, apitypes.LSXCmdSupported) {
+		op = "supported"
+		if dws, ok := d.(apitypes.StorageExecutorWithSupported); ok {
+			opResult, opErr := dws.Supported(ctx, store)
+			if opErr != nil {
+				err = opErr
+			} else {
+				result = opResult
+			}
+		} else {
+			err = apitypes.ErrNotImplemented
+		}
+	} else if strings.EqualFold(cmd, apitypes.LSXCmdInstanceID) {
 		op = "instance ID"
 		opResult, opErr := d.InstanceID(ctx, store)
 		if opErr != nil {
@@ -78,7 +89,7 @@ func Run() {
 			opResult.Driver = driverName
 			result = opResult
 		}
-	} else if cmd == "nextdevice" {
+	} else if strings.EqualFold(cmd, apitypes.LSXCmdNextDevice) {
 		op = "next device"
 		opResult, opErr := d.NextDevice(ctx, store)
 		if opErr != nil && opErr != apitypes.ErrNotImplemented {
@@ -86,7 +97,7 @@ func Run() {
 		} else {
 			result = opResult
 		}
-	} else if cmd == "localdevices" {
+	} else if strings.EqualFold(cmd, apitypes.LSXCmdLocalDevices) {
 		if len(args) < 4 {
 			printUsageAndExit()
 		}
@@ -101,7 +112,7 @@ func Run() {
 		} else {
 			result = opResult
 		}
-	} else if cmd == "wait" {
+	} else if strings.EqualFold(cmd, apitypes.LSXCmdWaitForDevice) {
 		if len(args) < 5 {
 			printUsageAndExit()
 		}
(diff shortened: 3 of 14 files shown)


codenrhoden:
Good improvement!

akutz:
Hi @codenrhoden,

Yeah, but I wonder if I should introduce a single function that Executors can implement called `Supported() bool` and if that returns false simply avoid any additional execution for that platform on a given client. It would also centralize the check for whether or not an executor is supported in the platform code to a single function instead of having executors implement a check in every function. I'm just not sure how some would do it, so I gave them the ability to do it at a functional level.

codenrhoden:
@akutz I was curious about that. It does seem beneficial to have a quick yes/no decision, and if support is "no", don't do anything else. But yeah -- how does an executor determine that? I think for each driver, there is probably a way. But perhaps the default is supported = yes, and if the executor has a way to determine it's *not* supported, then great.

ScaleIO can check if it has access to that sdc_guid binary, or that the `scini` kmod is loaded.
EBS/EFS could check that it is in fact on an EC2 instance by hitting the metadata server...

There are ways.

akutz:
Hi @codenrhoden,

ScaleIO is the easy one -- it can check for the existence of binaries (as you said). And EBS/EFS could hit the instance ID URI. 

I think this is why I made it up to the function to decide -- this way the executor assumes it is supported unless it isn't, you know? Then again, it could result in many unnecessary calls to `LocalDevices` from the client for unsupported platforms.

codenrhoden:
@akutz but in any scenario where an executor is *not* supported, it's a gain/savings over today. Of course, I don't know how much effort it is for you to add a `Supported()` method, but I do think it is a good thing. Especially looking down the road where we desire to have libStorage configured with multiple backends. The scenario that you've highlighted here could be a common one.

akutz:
Hi @codenrhoden,

Agreed. Care to help with this? I can take EBS and SIO if you want to tackle the others? I have to work out the core bits too, hence the unbalanced driver work. This is what the function will look like:

```go
// StorageExecutorWithSupported is an interface that executor implementations
// may use by defining the function "Supported(Context, Store) (bool, error)".
// This function indicates whether a storage platform is valid when executing
// the executor binary on a given client.
type StorageExecutorWithSupported interface {

	// Supported returns a flag indicating whether or not the platform
	// implementing the executor is valid for the host on which the executor
	// resides.
	Supported(ctx Context, opts Store) (bool, error)
}
```

codenrhoden:
@akutz Oh sure, take the easy ones. (Kidding!). Happy to help out.

akutz:
Hi @codenrhoden,

Absolutely I am! That's why I'm taking them. I have to also work out the framework bits :)

FWIW, I'm going to leave the existing code in this PR and just add to it. No reason it cannot support both, and this way for executors that elect to not implement the `Supported` function, they can still excerpt control over things via individual functions.

Also, just submit a PR per driver if you would. Don't submit them to this branch, submit the PRs to master. If they're complete by the time the core bits are ready I'll merge manually to the release branch. No reason to hold up the RR release for all drivers to support this today though.

akutz:
Hi @codenrhoden,

This isn't yet complete -- the client init logic must invoke the `Supported` function for the unique set of all discovered services' drivers and then cache that to prevent any future `InstanceID` or `LocalDevice` or Executor calls for that service if its driver isn't supported. So just some more work to do and then test. I wouldn't mind you reviewing what I've added though. Thanks!

api/types/types_drivers_executor.go
```diff
@@ -132,6 +158,7 @@ type ProvidesStorageExecutorCLI interface {
 // the driver implementations of the StorageExecutor interface.
 type StorageExecutorCLI interface {
 	StorageExecutorFunctions
+	StorageExecutorWithSupported
```
codenrhoden: One newbie question...

Why put `StorageExecutorWithSupported` here instead of putting `StorageExecutorFunctions` in `StorageExecutorWithSupported`? This seems to modify an existing interface, such that all executors will have to be updated to be usable. Is that true? If it is true and that is your intent, read no further...

The other way around would seeming make `StorageExecutorWithSupported` a new interface that you could check if an executor implemented before calling the `Supported` method -- thereby allowing old and new versions. I may be completely off here and speaking total nonesense.

akutz: Hi @codenrhoden,

Ah, good catch. It's the same reason I introduced the [`StorageDriverWithLogin`](https://github.com/emccode/libstorage/blame/master/api/types/types_drivers_storage.go#L160) interface in 6b6bd14fd3a9354951779be70af3620c909155af.

Let's look at how the `Login` function is [used](https://github.com/emccode/libstorage/blob/master/api/context/context.go#L127):

```go
// WithStorageSession returns a context that is logged into the storage
// platform.
func WithStorageSession(parent context.Context) (types.Context, error) {

	pctx := New(parent)

	d, ok := MustDriver(parent).(types.StorageDriverWithLogin)
	if !ok {
		pctx.Debug("driver is not StorageDriverWithLogin")
		return pctx, nil
	}

	sess, err := d.Login(pctx)
	if err != nil {
		pctx.WithError(err).Error("error logging into storage driver")
		return nil, err
	}
	if sess == nil {
		pctx.Warn("session is nil")
		return pctx, nil
	}

	return newContext(pctx, SessionKey, sess, nil, nil), nil
}
```

Please take note of this section of the above function:

```go
	d, ok := MustDriver(parent).(types.StorageDriverWithLogin)
	if !ok {
		pctx.Debug("driver is not StorageDriverWithLogin")
		return pctx, nil
	}
```

See how above I first check to see if the driver implements the interface `StorageDriverWithLogin` and I return early, not invoking the `Login` function if the provided driver does not have it?

This is a very clean manner in which Go makes it possible to make additive changes to code. Instead of adding the `Supported(Context, Store) (bool, error)` function to the existing `StorageExecutor` interface and breaking all the builds until that function is implemented for all the existing storage executor implementations, I simply created [a new interface](https://github.com/akutz/libstorage/blob/bugfix/missing-client-support/api/types/types_drivers_executor.go#L157-L173):

```go
// StorageExecutorWithSupported is an interface that executor implementations
// may use by defining the function "Supported(Context, Store) (bool, error)".
// This function indicates whether a storage platform is valid when executing
// the executor binary on a given client.
type StorageExecutorWithSupported interface {
	// Supported returns a flag indicating whether or not the platform
	// implementing the executor is valid for the host on which the executor
	// resides.
	Supported(
		ctx Context,
		opts Store) (bool, error)
}
```

So what does this mean? It means that the *existing* libStorage codebase continues to build and function as before without error. However, for those executors that *do* implement the interface I can [use that](https://github.com/akutz/libstorage/blob/bugfix/missing-client-support/cli/lsx/lsx.go#L71-L82):

```go
	if strings.EqualFold(cmd, apitypes.LSXCmdSupported) {
		op = "supported"
		if dws, ok := d.(apitypes.StorageExecutorWithSupported); ok {
			opResult, opErr := dws.Supported(ctx, store)
			if opErr != nil {
				err = opErr
			} else {
				result = opResult
			}
		} else {
			err = apitypes.ErrNotImplemented
		}
```

See above how we only invoke the `Supported` function if it is defined on that executor implementation? If it's not an `ErrNotImplemented` error is returned letting the eventual consumption of said error know that this executor simply hasn't yet implemented the new logic. And if it doesn't it will fall back to the first iteration of this PR where it's up to the individual executor functions to determine if something is supported.

akutz: By the way, your comment had it totally right:

> Why put `StorageExecutorWithSupported` here instead of putting `StorageExecutorFunctions` in `StorageExecutorWithSupported`? This seems to modify an existing interface, such that all executors will have to be updated to be usable. Is that true? If it is true and that is your intent, read no further...

> The other way around would seeming make `StorageExecutorWithSupported` a new interface that you could check if an executor implemented before calling the `Supported` method -- thereby allowing old and new versions. I may be completely off here and speaking total nonsense.

Golang refactoring used to *suck*. There is a stdlib [tool](https://godoc.org/golang.org/x/tools/refactor/rename) for it, but the last time I used it, well after the introduction of Go dependency vendoring, there were still an outpouring of grief that the tool would fail if vendored dependendencies were used. 

And since many Go tools that integrate with Atom use static analysis of the compiled packages, if the packages fail to compile you don't get things like intellisense, etc. Not to mention that after 10 errors Go just stops telling you the build errors.

So until the refactoring for Golang improves (I need to investigate to see if it has), I tend to make additive changes via new interfaces to prevent build failures.

akutz: @codenrhoden,

Ah, yes, another good question. These [two lines](https://github.com/akutz/libstorage/blob/bugfix/missing-client-support/api/types/types_drivers_executor.go#L160-L161) are your answer. Go can build cumulative interface definitions, but it cannot handle if the same function (or functions) exists in two of them.

So since the `StorageExecutorCLI` interface has a cumulative base of `StorageExecutorFunctions` and `StorageExecutorWithSupported`, if the latter inherited the former `StorageExecutorCLI` would not be able to "inherit" from them both. Try it and see what happens :)

image

```bash
[0]akutz@pax:types$ go build
# github.com/emccode/libstorage/api/types
./types_drivers_executor.go:175: duplicate method InstanceID
./types_drivers_executor.go:175: duplicate method LocalDevices
./types_drivers_executor.go:175: duplicate method NextDevice
```

akutz: And crap, I just realized I could use `StorageExecutorWithSupported` as the base of `StorageExecutorCLI` instead with the former using `StorageExecutorFunctions` as its base.

Dammit. Thanks @codenrhoden. I'll update the code. Good catch.

api/types/types_drivers_executor.go
```diff
@@ -131,7 +159,7 @@ type ProvidesStorageExecutorCLI interface {
 // StorageExecutorCLI provides a way to interact with the CLI tool built with
 // the driver implementations of the StorageExecutor interface.
 type StorageExecutorCLI interface {
-	StorageExecutorFunctions
+	StorageExecutorWithSupported
```
akutz: Here you go @codenrhoden, as requested :)

api/types/types_drivers_executor.go
```diff
@@ -122,6 +135,21 @@ type StorageExecutorFunctions interface {
 		opts *LocalDevicesOpts) (*LocalDevices, error)
 }
 
+// StorageExecutorWithSupported is an interface that executor implementations
+// may use by defining the function "Supported(Context, Store) (bool, error)".
+// This function indicates whether a storage platform is valid when executing
+// the executor binary on a given client.
+type StorageExecutorWithSupported interface {
+	StorageExecutorFunctions
```
akutz: Here you go @codenrhoden, as requested :)

api/types/types_drivers_executor.go
codenrhoden: Hurray, I get it!  Thanks a bunch for taking the time to explain that.

And, I didn't expect you to change anything. :)

However, I do like it better. :) Doesn't break anything as far as backwards compatibility goes, and now `StorageExecutorWithSupported` feels like a superset/enhancement.

Thanks again!

PR #297, opened 2016-10-12
isilon: add Supported() to executor

codenrhoden:
Isilon clients NFS mount the volumes, so make sure that mount.nfs
is present.

FWIW, I was hesitant to implement this way, as it is Linux specific. At first I was just going to check if `/usr/sbin/mount.nfs` existed, but that could be distro specific. But the executor already parses files in `/proc`, which binds it to Linux. So I'm not adding any new restrictions.

Also, it seemed possible that a client could by using NFSv4, and if they were they would need `mount.nfs4`, but the Linux driver actually doesn't use `mount -t nfs` or `mount -t nfs4`, it just calls `mount`, and can't be overriden, so this is also not a current issue. It merely relies on the presence of a `:` in the device to infer it is NFS.

This was inspired by #296

### DIFF
--- a/drivers/storage/isilon/executor/isilon_executor.go
+++ b/drivers/storage/isilon/executor/isilon_executor.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/akutz/gofig"
 	"github.com/akutz/goof"
+	"github.com/akutz/gotil"
 
 	"github.com/emccode/libstorage/api/registry"
 	"github.com/emccode/libstorage/api/types"
@@ -41,6 +42,11 @@ func (d *driver) Init(ctx types.Context, config gofig.Config) error {
 
 func (d *driver) Name() string {
 	return isilon.Name
+}
+
+func (d *driver) Supported(ctx types.Context, opts types.Store) (bool, error) {
+	// make sure NFS mounts can be done
+	return gotil.FileExistsInPath("mount.nfs"), nil
 }
 
 // InstanceID returns the local instance ID for the test


drivers/storage/isilon/executor/isilon_executor.go
```diff
@@ -43,6 +44,18 @@ func (d *driver) Name() string {
 	return isilon.Name
 }
 
+// Supported returns a flag indicating whether or not the platform
+// implementing the executor is valid for the host on which the executor
+// resides.
+func Supported(ctx types.Context, opts types.Store) (bool, error) {
+	// make sure mount.nfs is present
+	if err := exec.Command(
```
akutz: This assumes that both the `which` and `mount.nfs` commands are in the `PATH` variable as provided to the executor. Is there perhaps a cleaner way to do this?

codenrhoden: It is definitely assuming that. My (flawed?) reasoning for why this is okay is that libstorage is already making such assumptions for NFS: https://github.com/emccode/libstorage/blob/master/drivers/os/linux/linux.go#L185

Another tactic would be to survey our supported OS's, and see where the binary would be located on each of the, and then just `stat` for the file in all of the locations, returning `true` if any of them were successful.

akutz: Your logic works for me :) You could also use [this](https://github.com/akutz/gotil/blob/master/gotil.go#L165).

codenrhoden: Oh, nice! `LookPath()` is very convenient. It's `which` in Go. :)

I'm happy to change it. Do you prefer me to use `LookPath()` directly, or to import `gotil` ?

PR #312, opened 2016-10-31
Updated APIB Documentation

akutz:
This patch updates the libStorage APIB documentation. The updates included in this patch are available at the libStorage APIB documentation staging URL http://docs.libstoragedev.apiary.io.

I'm requesting reviews from @vladimirvivien, @codenrhoden, @mhrabovcin, and @kasisnu. Thanks!

Please note the main changes are in the headers section and around the `attachments` parameter.


libstorage.apib
```diff
@@ -468,33 +531,50 @@ Internal server error
...
+
+        A value that indicates how and if volume attachments should be retrieved.
+        <br/><br/>
+        The `attachments` parameter can be any of the following values:
+        <br/>
+        Value | Description
+        ------|------------
+        0 | No attachment information is requested. This is the default value and the same as omitting this parameter altogether.
+        1 | Attachment information is requested. This value indicates attachment information should be provided for all volumes.
+        2 | _My_ attachment information is requested. This value indicates attachment information should be provided for volumes attached to the instance specified in the [instance ID](#def-instance-id) request header should be returned. If this bit is set then the instance ID header is required.
```
codenrhoden: The "should be provided" .... "should be returned" part reads funny. Perhaps rewrite as:

> This value indicates attachment information should be returned for volumes attached to the instance specified in the [instance ID](#def-instance-id) request header. If this bit is set then the instance ID header is required.

libstorage.apib
```diff
@@ -1010,34 +1107,59 @@ Internal server error
...
+        4 | Attempt to map devices provided via the [local devices](#def-local-devices) request header to the appropriate attachment information. If this bit is set then the instance ID and local device headers are required.
```
codenrhoden, line 1137: Will it be up to the driver to know when these values are appropriate? Values `2` and `4` could be weird in this scenario -- if the client is requesting info about a volume that isn't attached to itself, but you add flags that try to map or otherwise require info from the client.

libstorage.apib
```diff
@@ -468,33 +531,50 @@ Internal server error
...
```
codenrhoden, line 553: After reading through the entire PR, I have one outstanding question in this area, combining the local devices header and the `GET /volumes?{attachments}` route...

My understanding is that the LocalDevices header is service/driver specific. E.G. the client provides a list of ScaleIO devices to the ScaleIO driver, EBS devices to the EBS driver, etc. But for `GET /volumes`, there isn't a specific service involved. So what does a client do in that case? Provide a full device map of everything attached?

If _that_ was/is the case, it seems better to me to have it built into the libstorage client (rather than an executor) to grab a generic version of this, and then have the storage drivers themselves parse out what is relevant to them (e.g. ScaleIO knows to look at `/dev/scini*`. Not sure if this possible or not (There may be drivers that overlap when it comes to device names?). It seems difficult to handle NFS, iSCSI, kernel block devices, etc. in a generic way, but if this call needs all that info...

Hope what I'm asking makes sense.

akutz: Hi @codenrhoden,

Great question. That's the reason the `Libstorage-Instanceid` and `Libstorage-Localdevices` headers start with `DRIVER` -- it indicates the type of driver for which the data is applicable. 

### Getting Volumes for A Single Service

Imagine the following HTTP request:

```
GET /volumes/ebs00

Libstorage-Instanceid: ebs=i-2348,region%3Dwest
Libstorage-Instanceid: scaleio=4f580-3091
Libstorage-Localdevices: ebs=/dev/xvdf::/dev/xvdf
Libstorage-Instanceid: scaleio=vol-123::/dev/scini1,vol-349::/dev/scini2
```

The above request is requesting all volumes for service `ebs00`. Let's assume this service is configured with the `ebs` driver. The request would trigger the [following](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume.go#L62-L70) route:

``` go
        // get all volumes from a specific service
        httputils.NewGetRoute(
            "volumesForService",
            "/volumes/{service}",
            r.volumesForService,
            handlers.NewServiceValidator(),
            handlers.NewStorageSessionHandler(),
            handlers.NewSchemaValidator(nil, schema.VolumeMapSchema, nil),
        ),
```

As you can see, the above route uses the handler returned by the `NewServiceValidator()` call. Thus the above request would be passed through [this](https://github.com/codedellemc/libstorage/blob/master/api/server/handlers/handlers_service_validator.go#L32-L51) handler:

``` go
// Handle is the type's Handler function.
func (h *serviceValidator) Handle(
    ctx types.Context,
    w http.ResponseWriter,
    req *http.Request,
    store types.Store) error {

    if !store.IsSet("service") {
        return utils.NewStoreKeyErr("service")
    }

    serviceName := store.GetString("service")
    service := services.GetStorageService(ctx, serviceName)
    if service == nil {
        return utils.NewNotFoundError(serviceName)
    }

    ctx = context.WithStorageService(ctx, service)
    return h.handler(ctx, w, req, store)
}
```

The following [line](https://github.com/codedellemc/libstorage/blob/master/api/server/handlers/handlers_service_validator.go#L49) from the above function takes the retrieved storage service and inserts into the request context:

``` go
ctx = context.WithStorageService(ctx, service)
```

The `context.WithStorageService` function, defined [here](https://github.com/codedellemc/libstorage/blob/master/api/context/context.go#L91-L119), looks like this:

``` go
// WithStorageService returns a new context with the StorageService as the
// value and attempts to assign the service's associated InstanceID and
// LocalDevices (by way of the service's StorageDriver) to the context as well.
func WithStorageService(
    parent context.Context, service types.StorageService) types.Context {

    var (
        driver     = service.Driver()
        driverName = strings.ToLower(driver.Name())
    )

    parent = newContext(parent, DriverKey, driver, nil, nil)

    // set the service's InstanceID if present
    if iidm, ok := parent.Value(AllInstanceIDsKey).(types.InstanceIDMap); ok {
        if iid, ok := iidm[driverName]; ok {
            parent = newContext(parent, InstanceIDKey, iid, nil, nil)
        }
    }

    // set the service's LocalDevices if present
    if ldm, ok := parent.Value(AllLocalDevicesKey).(types.LocalDevicesMap); ok {
        if ld, ok := ldm[driverName]; ok {
            parent = newContext(parent, LocalDevicesKey, ld, nil, nil)
        }
    }

    return newContext(parent, ServiceKey, service, nil, nil)
}
```

As is illustrated above, the `WithStorageService` function returns a new context with the instance ID and local devices map injected into the request context that map to the driver for which the storage service is configured. In this case the `ebs00` service uses the `ebs` driver, so the following headers from the request come into play:

```
Libstorage-Instanceid: ebs=i-2348,region%3Dwest
Libstorage-Localdevices: ebs=/dev/xvdf::/dev/xvdf
```

The instance ID and local devices headers for ScaleIO were simply ignored for this service.

Then, inside the driver's `Volumes` function, if the driver needs to know the instance ID or local devices, the driver can simply invoke one of the following functions:
- [`context.InstanceID(context.Context) (*types.InstanceID, bool)`](https://github.com/codedellemc/libstorage/blob/master/api/context/context.go#L297)
- [`context.MustInstanceID(context.Context) *types.InstanceID`](https://github.com/codedellemc/libstorage/blob/master/api/context/context.go#L304)
- [`context.LocalDevices(context.Context) (*types.LocalDevices, bool)`](https://github.com/codedellemc/libstorage/blob/master/api/context/context.go#L310)

The above functions are simply variations of helpers that ultimately invoke `context.Value(key interface{}) interface{}`. Both the keys for retrieving or setting the [instance ID](https://github.com/codedellemc/libstorage/blob/master/api/context/context_keys.go#L50) and [local devices](https://github.com/codedellemc/libstorage/blob/master/api/context/context_keys.go#L27) are defined in [`context_keys.go`](https://github.com/codedellemc/libstorage/blob/master/api/context/context_keys.go).

### Getting Volumes for Multiple Services

The previous section described the process of how instance ID and local device headers are transported, injected into the context, and used by drivers when getting volumes or a volume for a single service. This section describes the steps taken when getting volumes for multiple services:

Imagine the following HTTP request:

```
GET /volumes

Libstorage-Instanceid: ebs=i-2348,region%3Dwest
Libstorage-Instanceid: scaleio=4f580-3091
Libstorage-Localdevices: ebs=/dev/xvdf::/dev/xvdf
Libstorage-Instanceid: scaleio=vol-123::/dev/scini1,vol-349::/dev/scini2
```

The above request would use the following [route](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume.go#L54-L60):

``` go
        // get all volumes from all services
        httputils.NewGetRoute(
            "volumes",
            "/volumes",
            r.volumes,
            handlers.NewSchemaValidator(nil, schema.ServiceVolumeMapSchema, nil),
        ),
```

Notice the above route has no service handler. That's because the above route does not contain any information with which to determine the service. But before a call reaches a driver implementation the context is still populated with instance ID and local device information just like in the previous section. Here's [how](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L43-L62):

``` go
    for service := range services.StorageServices(ctx) {

        run := func(
            ctx types.Context,

            ctx = context.WithStorageService(ctx, svc)

            var err error
            if ctx, err = context.WithStorageSession(ctx); err != nil {
                return nil, err
            }

            return getFilteredVolumes(ctx, req, store, svc, opts, filter)
        }

        task := service.TaskExecute(ctx, run, schema.VolumeMapSchema)
        taskIDs = append(taskIDs, task.ID)
        tasks[service.Name()] = task
    }
```

The above code is executed when a call is routed from `/volumes`. All configured services are enumerated, and for each service the context is updated with this [call](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L49):

``` go
ctx = context.WithStorageService(ctx, svc)
```

Please note this is the same function call in the service handler that is absent from this route. Since the route itself doesn't inform the service, once inside the code that handles the route, each service is enumerated and before a call to its driver is invoked to get volumes, the same logic that is executed for a service-specific route is executed on the service here as well.

Hopefully this helps explain some of the questions you had.

akutz: Hi @codenrhoden,

> But for GET /volumes, there isn't a specific service involved. So what does a client do in that case? Provide a full device map of everything attached?

Yes. When the libStorage client is first initialized it asks the server for information about all configured services. From that a unique set of drivers is compiled and for each driver the executor is used to grab the driver's instance ID. This is cached for the duration of the client.

For local devices, any call that might require local devices causes the executor to gather that information. For service-specific calls the local devices are retrieved only for the service's driver type. For global calls all local device maps are obtained -- one for each driver type.

libstorage.apib
akutz: Yes, it's up to the driver. The driver can retrieve all volumes and all attachment information all the time and the framework will filter the data based on the mask. However, if the driver can more optimally perform the volume and attachment retrieval then it should.

libstorage.apib
codenrhoden: It absolutely does help explain. Thank you!

A major piece that I was missing is that in the `GET /volumes` case, the client can just send the header multiple times for each service. I remember reading that now, but had forgotten it. The hurdles of coming up to speed on a new project. :)

Thanks as always for your detailed response, with great examples!

libstorage.apib
akutz: Thank you @codenrhoden. The PR has been updated with 4d49ca40fae0e98418d8b6559623197cc25f881c based on your feedback.

libstorage.apib
akutz: > the client can just send the header multiple times for each service

Technically the client sends the header multiple times, one per unique driver type. For example, imagine a libStorage server configured as such:

``` yaml
libstorage:
  server:
    services:
      ebs00:
        driver: ebs
      ebs01:
        driver: ebs
      ebs02:
        driver: ebs
      sio00:
        driver: scaleio
      sio01:
        driver: scaleio
```

The above libStorage configuration defines five services: three using the EBS driver and two using the ScaleIO driver. However, when the libStorage client sends the server instance ID and local device headers, this is what will be sent:

```
Libstorage-Instanceid: ebs=i-2348,region%3Dwest
Libstorage-Instanceid: scaleio=4f580-3091
Libstorage-Localdevices: ebs=/dev/xvdf::/dev/xvdf
Libstorage-Instanceid: scaleio=vol-123::/dev/scini1,vol-349::/dev/scini2
```

Even though the server has three configured services that use the EBS driver and two configured services that use the ScaleIO driver, the client only needs to send the instance ID and local devices for EBS once. The same for ScaleIO. 

The use of multiple headers correlates to unique drivers for configured services, not for the number of configured services themselves.

Does that make sense?

akutz: @codenrhoden, does the above clarification make sense?

codenrhoden: Oh it does, yes. And is so obvious/clear I didn't bother to reply. :) Sorry.

I was careless with my usage of "driver" vs "service" in this case. Didn't really think about it, but will do so in the future.

libstorage.apib
```diff
@@ -468,33 +531,50 @@ Internal server error
...
 
 + Parameters
 
-    + attachments (required)
-
-        A flag that indicates that volume attachments should be retrieved.
-        If this flag is provided then the `LibStorage-InstanceID` header must
-        be set -- a comma-delimited list of service name/instance ID,
-        colon-delimited pairs.
+    + attachments (string,optional)
```
mhrabovcin, line 542: Why is this `string` type when values listed down are `[1-8]`?

akutz: Hi @mhrabovcin,

Because query string params are all either string or boolean. Additionally I wanted the param to be backwards compatible when it was a boolean flag, so if a truthy value is passed I can interpret it on the backend like so:

``` go
// ParseVolumeAttachmentTypes parses a value into a VolumeAttachmentsTypes
// value.
func ParseVolumeAttachmentTypes(v interface{}) VolumeAttachmentsTypes {
    switch tv := v.(type) {
    case VolumeAttachmentsTypes:
        return tv
    case bool:
        if tv {
            return VolumeAttachmentsTypes(14)
        } else {
            return VolumeAttachmentsNone
        }
    case int:
        return VolumeAttachmentsTypes(tv)
    case uint:
        return VolumeAttachmentsTypes(tv)
    case int8:
        return VolumeAttachmentsTypes(tv)
    case uint8:
        return VolumeAttachmentsTypes(tv)
    case int16:
        return VolumeAttachmentsTypes(tv)
    case uint16:
        return VolumeAttachmentsTypes(tv)
    case int32:
        return VolumeAttachmentsTypes(tv)
    case uint32:
        return VolumeAttachmentsTypes(tv)
    case int64:
        return VolumeAttachmentsTypes(tv)
    case uint64:
        return VolumeAttachmentsTypes(tv)
    case string:
        if i, err := strconv.ParseInt(tv, 10, 64); err == nil {
            return ParseVolumeAttachmentTypes(i)
        }
        if b, err := strconv.ParseBool(tv); err == nil {
            return ParseVolumeAttachmentTypes(b)
        }
        return VolumeAttachmentsNone
    }
    return VolumeAttachmentsNone
}
```

akutz:
I'm closing this PR in favor or merging the doc commits into PR #313 that includes the code to update the `attachments` parameter.

PR #314, opened 2016-11-02
Missing AWS Config Support for EBS driver

akutz:
This patch reintroduces support for the "aws" config namespace that used to exist in pre-0.4 versions of REX-Ray. For example:

```yaml
aws:
  accessKey: 123
  secretKey: 456
```

The above is now valid once again. This PR handles REX-Ray issue [\#617](https://github.com/codedellemc/rexray/issues/617). Many thanks to @kbeckman for discovering the issue.

### DIFF
--- a/drivers/storage/ebs/ebs.go
+++ b/drivers/storage/ebs/ebs.go
@@ -8,8 +8,11 @@ const (
 	// Name is the provider's name.
 	Name = "ebs"
 
-	// OldName is the provider's old name.
-	OldName = "ec2"
+	// NameEC2 is the provider's old EC2 name.
+	NameEC2 = "ec2"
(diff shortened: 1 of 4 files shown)


akutz:
I validated this using the EBS test Vagrantfile and following config:

``` yaml
rexray:
  logLevel: debug
libstorage:
  host:     tcp://127.0.0.1:7979
  embedded: true
  service:  ebs
  server:
    endpoints:
      public:
        address: tcp://:7979
    services:
      ebs:
        driver: ebs
aws:
  accessKey: 123
  secretKey: 456
```

Listing volumes now works, even with a config that uses the `aws` key for configuring the EBS driver.

PR #319, opened 2016-11-04
Volume Attachment States

akutz:
This patch adds the AttachmentState field to the Volume object model to more summarily reflect a volume's attachment state.

### DIFF
--- a/api/server/router/volume/volume_routes.go
+++ b/api/server/router/volume/volume_routes.go
@@ -395,6 +395,9 @@ func (r *router) volumeCreate(
 			}
 		}
 
+		if v.AttachmentState == 0 {
+			v.AttachmentState = types.VolumeAvailable
+		}
 		return v, nil
 	}
 
@@ -439,6 +442,9 @@ func (r *router) volumeCopy(
 			}
 		}
 
+		if v.AttachmentState == 0 {
+			v.AttachmentState = types.VolumeAvailable
+		}
 		return v, nil
 	}
 
@@ -517,6 +523,10 @@ func (r *router) volumeAttach(
 			}
 		}
 
+		if v.AttachmentState == 0 {
+			v.AttachmentState = types.VolumeAttached
+		}
+
 		return &types.VolumeAttachResponse{
 			Volume:      v,
 			AttachToken: attTokn,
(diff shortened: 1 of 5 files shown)


api/server/router/volume/volume_routes.go
```diff
@@ -517,6 +523,10 @@ func (r *router) volumeAttach(
 			}
 		}
 
+		if v.AttachmentState == 0 {
+			v.AttachmentState = types.VolumeAttached
+		}
+
```
codenrhoden, line 529: why is this one `0 -> attached` and all the others `0 -> available`? My guess is that this is basically setting the default value, and if a call to `Attach()` is returning without an error, it can be assumed that the volume is attached.  makes sense, but want to double check.

api/types/types_model.go
```diff
@@ -139,28 +169,28 @@ type Volume struct {
 	IOPS int64 `json:"iops,omitempty" yaml:"iops,omitempty"`
 
 	// The name of the volume.
-	Name string `json:"name"`
+	Name string `json:"name" yaml:"name,omitempty"`
```
codenrhoden, line 172: Why is YAML `omitempty`, but not JSON? Seems like the only outlier?

codenrhoden (commented):
I know this is already merged, but a couple simple questions for you...

api/server/router/volume/volume_routes.go
akutz: You inferred the reasoning correctly.

akutz:
Oh, because I got Copy/Paste happy with the YAML update. Doesn't matter ultimately because the YAML tags were originally there to influence the CLI output which is no longer based on YAML.

api/types/types_model.go
akutz: Oh, because I got Copy/Paste happy with the YAML update. Doesn't matter ultimately because the YAML tags were originally there to influence the CLI output which is no longer based on YAML.

api/server/router/volume/volume_routes.go
akutz: Although now that you point this out, I need to update the above to the following logic:

```
if v.AttachmentState == 0 && len(v.Attachments) > 0 {
    if "attached to the same instance as the instance ID in the context" {
        v.AttachmentState = types.VolumeAttached
    } else {
        v.AttachmentState = types.VolumeUnavailable
    }
}
```

akutz: Wait, no I won't. The API call asserts the instance ID must be included to do the attach op, which means when the volume returns the volume WILL be attached (if successful) to whichever instance was specified in the call.

PR #327, opened 2016-11-09
Volume Attachment Filtering Logic

akutz:
This patch updates the volume attachment filtering logic so that it can return both attached and unattached volumes in the same query if both bits are set in the mask. Additionally, if bit 2 is set it does not restrict the volumes being returned to only those attached to the provided instance unless the mask also includes bit 8.

Per [this conversation](https://github.com/codedellemc/rexray/pull/619#discussion_r86922502), I think the following helps to explain the purpose and reasoning behind the attachment flags and logic of how they are intended to function:

---

Hi @codenrhoden,

So I hear you and I understand. I think part of the issue is that I want to be able to do this, that, and the other with a single query. Meaning I'd like to be able to say "get all volumes and their attachments" or "get all volumes that are attached" or "get all volumes that are not attached" or "get all volumes that are attached to me and all volumes that are not attached".

What I don't want is for people to have to worry about is performing multiple queries to get what they want. Let's review the values here:

```
const (
	// VolumeAttachmentsRequested indicates attachment information is requested.
	VolumeAttachmentsRequested VolumeAttachmentsTypes = 1 << iota // 1

	// VolumeAttachmentsMine indicates attachment information should
	// be returned for volumes attached to the instance specified in the
	// instance ID request header. If this bit is set then the instance ID
	// header is required.
	VolumeAttachmentsMine // 2

	// VolumeAttachmentsDevices indicates an attempt should made to map devices
	// provided via the local devices request header to the appropriate
	// attachment information. If this bit is set then the instance ID and
	// local device headers are required.
	VolumeAttachmentsDevices // 4

	// VolumeAttachmentsAttached indicates only volumes that are attached
	// should be returned.
	VolumeAttachmentsAttached // 8

	// VolumeAttachmentsUnattached indicates only volumes that are unattached
	// should be returned.
	VolumeAttachmentsUnattached // 16
)
```

As I see it, here are the valid masks:

Mask | Components | Description
------|--------------|------------
`0`| `0` | The default value. This indicates no attachment information is requested.
`1` | `1` | Request attachment information for all retrieved volumes
`3` | `1\|2` | Request attachment information for volumes attached to the instance provided in the instance ID
`7` | `1\|2\|4` | Request attachment information for volumes attached to the instance provided in the instance ID and perform device mappings where possible.
`9` | `1\|8` | Request attachment information for all retrieved volumes and return only volumes that are attached to some instance.
`17` | `1\|16` | Request attachment information for all retrieved volumes and return only volumes that are not attached to any instance.

Here's where it gets a little more interesting:

Mask | Components | Description
------|--------------|------------
`11` | `1\|2\|8` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID.
`15` | `1\|2\|4\|8` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID and perform device mappings where possible.
`27` | `1\|2\|8\|16` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID *or* are not attached to any instance at all. tl;dr - _Attached To Me or Available_
`31` | `1\|2\|4\|8\|16` | Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID *or* are not attached to any instance at all and perform device mappings where possible. tl;dr - _Attached To Me With Device Mappings or Available_

---

Hi @codenrhoden,

I got cut off above (hit enter too soon). I may be totally off base above, but can you see what I was attempting with the above mask?

One thing to note, and in regards to something you said:

> Why use the bitmask at all, then, instead of just `Requested`?

Because to do any of the mapping to an instance or device mapping requires that clients send in the request headers the instance ID and local device information. The mask provides the ability to retrieve volumes without also requiring the presence of the instance ID and/or local device information in the headers.

Previously I assumed drivers would check to see if the instance ID was present, and if so, automatically return volume attachment information for only that instance. 

However that was too constricting as it didn't leave us the ability to create a client that automatically handles sending the instance ID for a user. Additionally, it meant for some drivers we could not request volumes that are both attached to a provided instance but also available (unattached to any other instance). 

See what I mean?

Or am I totally missing something obvious here? I could be. What are your thoughts?

FWIW, see [these tests](https://github.com/akutz/libstorage/blob/bugfix/attachment-filtering-logic/drivers/storage/vfs/tests/vfs_test.go#L166-L312) I added to showcase the various ways to use the mask. Each test asserts the return values are what they are expected to be.

---

Hi @codenrhoden,

One last thing -- I think the reason this is confusing is because I'm using a mask for what is ultimately two purposes:

1. Which data to retrieve
2. Filtering a separate, but related, data set

That is the mask is used to indicate which attachment information to retrieve, but also which volumes to return, filtered by the retrieved attachment data.

This is perhaps not the best way to use a mask -- we should ultimately have a true filtering capability. I was just trying to combine two things into one for now for the sake of simplicity since I have yet to fully implement the LDAP BNF search grammar I implemented for simplistic querying in libStorage [`#0730430`](https://github.com/codedellemc/libstorage/commit/0730430da9d9009dd6ff6d330cf706908b5c7d0f).

For more information on the LDAP search syntax see [RFC 4515](https://tools.ietf.org/search/rfc4515) and this [article](http://matt.might.net/articles/grammars-bnf-ebnf/) on Backus-Naur Form (BNF).

---

Hi @codenrhoden,

I just noticed this after re-reading your comments:

> This makes sense on its own, but gives Mine precedence over Attached and Unattached.

Well, `mine` has no precedence over `unattached` as `mine` only has any effect if a volume has attachments.

And yes, `mine` has precedence over `attached`, but if you take a close look at the tables above you'll notice there is no attempt of the following:

> Request attachment information for all retrieved volumes and return only volumes that attached to the instance provided in the instance ID or are not attached to any instance at all or **are attached at all**.

The above would be what you described -- a mask that's the same as saying get all attachments for all volumes, right?

As it is, the logic is such to enable requesting all volumes attached to the provided instance or all volumes not attached to any instance, aka. volumes available to be attached.

The order of the server logic is specifically designed to enable this. That's why the `attached` bit only removes unattached volumes if the `unattached` bit is not set. That's why the `unattached` bit only removed attached volumes if the`attached` bit is not set.

Again, I do have [these tests](https://github.com/akutz/libstorage/blob/bugfix/attachment-filtering-logic/drivers/storage/vfs/tests/vfs_test.go#L166-L312) that validate the desired logic. 

I'm fully willing to admit I'm off base on all of this, but I also think it's possible I just did a **supremely** shitty job describing the purpose and implementation of these things with which to begin, and it's lead to much confusion.

### DIFF
--- a/api/server/handlers/handlers_query_params.go
+++ b/api/server/handlers/handlers_query_params.go
@@ -49,7 +49,9 @@ func (h *queryParamsHandler) Handle(
 			if len(v[0]) == 0 {
 				store.Set(k, true)
 			} else {
-				if b, err := strconv.ParseBool(v[0]); err == nil {
+				if i, err := strconv.ParseInt(v[0], 10, 64); err == nil {
+					store.Set(k, i)
+				} else if b, err := strconv.ParseBool(v[0]); err == nil {
 					store.Set(k, b)
 				} else {
 					store.Set(k, v[0])
--- a/api/server/router/volume/volume_routes.go
+++ b/api/server/router/volume/volume_routes.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"sync"
 
+	log "github.com/Sirupsen/logrus"
 	"github.com/akutz/goof"
 
 	"github.com/codedellemc/libstorage/api/context"
@@ -127,6 +128,66 @@ func (r *router) volumesForService(
 		http.StatusOK)
 }
 
+func handleVolAttachments(
+	ctx types.Context,
+	lf log.Fields,
+	iid *types.InstanceID,
+	vol *types.Volume,
+	attachments types.VolumeAttachmentsTypes) bool {
+
+	if attachments == 0 {
+		vol.Attachments = nil
+		return true
+	}
+
+	if lf == nil {
+		lf = log.Fields{}
+	}
+
+	// if only the requesting instance's attachments are requested then
+	// filter the volume's attachments list
+	if attachments.Mine() {
+		atts := []*types.VolumeAttachment{}
+		for _, a := range vol.Attachments {
+			alf := log.Fields{
+				"attDeviceName": a.DeviceName,
+				"attDountPoint": a.MountPoint,
+				"attVolumeID":   a.VolumeID,
+			}
+			if strings.EqualFold(iid.ID, a.InstanceID.ID) {
+				atts = append(atts, a)
+				ctx.WithFields(lf).WithFields(alf).Debug(
+					"including volume attachment")
+			} else {
+				ctx.WithFields(lf).WithFields(alf).Debug(
+					"omitting volume attachment")
+			}
+		}
+		vol.Attachments = atts
+		ctx.WithFields(lf).Debug("included volume attached to instance")
+	}
+
+	// if the volume has no attachments and the mask indicates that
+	// only attached volumes should be returned then omit this volume
+	if len(vol.Attachments) == 0 &&
+		attachments.Attached() &&
+		!attachments.Unattached() {
+		ctx.WithFields(lf).Debug("omitting unattached volume")
+		return false
+	}
+
+	// if the volume has attachments and the mask indicates that
+	// only unattached volumes should be returned then omit this volume
+	if len(vol.Attachments) > 0 &&
+		!attachments.Attached() &&
+		attachments.Unattached() {
+		ctx.WithFields(lf).Debug("omitting attached volume")
+		return false
+	}
+
+	return true
+}
+
 func getFilteredVolumes(
 	ctx types.Context,
 	req *http.Request,
@@ -147,6 +208,8 @@ func getFilteredVolumes(
 		return nil, utils.NewMissingInstanceIDError(storSvc.Name())
 	}
 
+	ctx.WithField("attachments", opts.Attachments).Debug("querying volumes")
+
 	objs, err := storSvc.Driver().Volumes(ctx, opts)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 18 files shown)


akutz:
Hi @codenrhoden,

I added the following helper masks to hopefully make things easier to understand:

``` go
const (
    // VolAttNone is the default value. This indicates no attachment
    // information is requested.
    VolAttNone VolumeAttachmentsTypes = 0

    // VolAttFalse is an alias for VolAttNone.
    VolAttFalse = VolAttNone

    // VolAttReq requests attachment information for all retrieved volumes.
    //
    // Mask: 1
    VolAttReq = VolumeAttachmentsRequested

    // VolAttReqForInstance requests attachment information for volumes attached
    // to the instance provided in the instance ID
    //
    // Mask: 1 | 2
    VolAttReqForInstance = VolAttReq | VolumeAttachmentsMine

    // VolAttReqWithDevMapForInstance requests attachment information for
    // volumes attached to the instance provided in the instance ID and perform
    // device mappings where possible.
    //
    // Mask: 1 | 2 | 4
    VolAttReqWithDevMapForInstance = VolAttReqForInstance |
        VolumeAttachmentsDevices

    // VolAttReqOnlyAttachedVols requests attachment information for all
    // retrieved volumes and return only volumes that are attached to some
    // instance.
    //
    // Mask: 1 | 8
    VolAttReqOnlyAttachedVols = VolAttReq | VolumeAttachmentsAttached

    // VolAttReqOnlyUnattachedVols requests attachment information for
    // all retrieved volumes and return only volumes that are not attached to
    // any instance.
    //
    // Mask: 1 | 16
    VolAttReqOnlyUnattachedVols = VolAttReq | VolumeAttachmentsUnattached

    // VolAttReqOnlyVolsAttachedToInstance requests attachment
    // information for all retrieved volumes and return only volumes that
    // attached to the instance provided in the instance ID.
    //
    // Mask: 1 | 2 | 8
    VolAttReqOnlyVolsAttachedToInstance = VolAttReqForInstance |
        VolumeAttachmentsAttached

    // VolAttReqWithDevMapOnlyVolsAttachedToInstance requests attachment
    // information for all retrieved volumes and return only volumes that
    // attached to the instance provided in the instance ID and perform device
    // mappings where possible.
    //
    // Mask: 1 | 2 | 4 | 8
    VolAttReqWithDevMapOnlyVolsAttachedToInstance = VolumeAttachmentsDevices |
        VolAttReqOnlyVolsAttachedToInstance

    // VolAttReqTrue is an alias for
    // VolAttReqWithDevMapOnlyVolsAttachedToInstance.
    VolAttReqTrue = VolAttReqWithDevMapOnlyVolsAttachedToInstance

    // VolAttReqOnlyVolsAttachedToInstanceOrUnattachedVols requests attachment
    // information for all retrieved volumes and return only volumes that
    // attached to the instance provided in the instance ID or are not attached
    // to any instance at all. tl;dr - Attached To Me or Available
    //
    // Mask: 1 | 2 | 8 | 16
    VolAttReqOnlyVolsAttachedToInstanceOrUnattachedVols = 0 |
        VolAttReqOnlyVolsAttachedToInstance |
        VolumeAttachmentsUnattached

    // VolAttReqWithDevMapOnlyVolsAttachedToInstanceOrUnattachedVols requests
    // attachment information for all retrieved volumes and return only volumes
    // that attached to the instance provided in the instance ID or are not
    // attached to any instance at all and perform device mappings where
    // possible. tl;dr - Attached To Me With Device Mappings or Available
    //
    // Mask: 1 | 2 | 4 | 8 | 16
    VolAttReqWithDevMapOnlyVolsAttachedToInstanceOrUnattachedVols = 0 |
        VolumeAttachmentsDevices |
        VolAttReqOnlyVolsAttachedToInstanceOrUnattachedVols
)
```

akutz:
Hi @codenrhoden,

Please note that I updated the above comment to reflect the refinement of the common masks.

akutz:
FYI, this PR has a merge conflict, but it's a tiny one. It's on line 159 of `api/types/types_drivers_storage.go` and looks like this:

```diff
<<<<<<< HEAD
			return VolumeAttachmentsTrue
=======
			return VolAttReqTrue
>>>>>>> bugfix/attachment-filtering-logic
```

I will resolve the merge conflict manually.

PR #329, opened 2016-11-11
Fix for Absent Attachment State

akutz:
This patch fixes the missing attachment state when listing volumes.


codenrhoden (commented):
Just wondering if you should set:
`vol.AttachmentState = types.VolumeAttachmentStateUnknown` up at Line 140.  That way it's always initialized, even in the early return case.

PR #341, opened 2016-11-16
Added Example for EC2 Vagrantfile

akutz:
This patch adds an example for bringing a libStorage environment online
with REX-Ray using EC2, EBS, and EFS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,7 +23,9 @@ cli/semaphores/signal
 cli/semaphores/unlink
 api/api_generated.go
 api/server/executors/executors_generated.go
-drivers/storage/ebs/tests/.vagrant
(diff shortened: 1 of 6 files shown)


codenrhoden:
@akutz - I know we chatted a bit about it previously, but for posterity, what is the difference between the files in this `example` folder, and the files found in the driver's `test` folder?

akutz:
Hi @codenrhoden,

[This](https://youtu.be/gqQ99s4Ywnw) clip comes to mind when you comment on merged issues :)

FWIW, the difference is that `examples/ec2/vagrant/Vagrantfile` sets up a server with both EBS and EFS services configured as the example is provider agnostic.

PR #343, opened 2016-11-18
EFS Updates

akutz:
This patch updates EFS to use the Login interface for the storage driver as well as calculate the full instance ID on the client-side. Additionally, the new configuration keys `efs.endpoint` and `efs.endpointFormat` are introduced to enable custom endpoint URLs outside the scope of `*.amazonaws.com` per this [REX-Ray PR/issue](https://github.com/codedellemc/rexray/pull/625). 

A subsequent update will occur to add the new `endpointFormat` key to the EBS driver as well. 

This is a WiP PR. I'd appreciate it if @codenrhoden, @mhrabovcin, and @kasisnu could all please take a look and share their thoughts.

### DIFF
--- a/drivers/storage/efs/storage/efs_storage.go
+++ b/drivers/storage/efs/storage/efs_storage.go
@@ -3,7 +3,9 @@ package storage
 import (
 	"crypto/md5"
 	"fmt"
+	"hash"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -16,6 +18,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/credentials/ec2rolecreds"
 	"github.com/aws/aws-sdk-go/aws/ec2metadata"
 	"github.com/aws/aws-sdk-go/aws/session"
+	awsec2 "github.com/aws/aws-sdk-go/service/ec2"
 	awsefs "github.com/aws/aws-sdk-go/service/efs"
 
 	"github.com/codedellemc/libstorage/api/context"
@@ -30,8 +33,17 @@ const (
 
 // Driver represents a EFS driver implementation of StorageDriver
 type driver struct {
-	config   gofig.Config
-	awsCreds *credentials.Credentials
+	config              gofig.Config
+	region              *string
+	endpoint            *string
+	endpointFormat      string
+	endpointEC2         *string
+	endpointEC2Format   string
+	maxRetries          *int
+	tag                 string
+	accessKey           string
+	secGroups           []string
+	disableSessionCache bool
 }
 
 func init() {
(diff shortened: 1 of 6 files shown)


codenrhoden (commented):
I'm not familiar enough with `EFS` to see any lingering errors, but the changes look good and bring this in line with `EBS`. The `EndpointFormat` is a nice enhancement.

mhrabovcin:
I don't understand why is this driver adding API connection to EC2 service and not using it anywhere? If we want to reuse AWS service connection abstraction between different drivers I'd say it would be worth extracting this to a separate package and not have it in EFS driver.

drivers/storage/efs/storage/efs_storage.go
```diff
@@ -50,37 +61,210 @@ func (d *driver) Name() string {
...
+
+func (a *awsLogger) Log(args ...interface{}) {
+	a.ctx.Infoln(args...)
+}
+
+func mustSession(ctx types.Context) *awsService {
+	return context.MustSession(ctx).(*awsService)
+}
+
+func mustInstanceIDID(ctx types.Context) *string {
```
mhrabovcin, line 283: Is this a typo?

drivers/storage/efs/storage/efs_storage.go
```diff
@@ -604,51 +796,43 @@ func (d *driver) getVolumeAttachments(
...
-			WithLogger(newAwsLogger()).
-			WithLogLevel(aws.LogDebug)
-	}
+// Retrieve config arguments
+func (d *driver) getAccessKey() string {
+	return d.config.GetString(efs.ConfigEFSAccessKey)
+}
 
-	return awsefs.New(session.New(), config)
+func (d *driver) secretKey() string {
```
mhrabovcin: This should be probably `getSecretKey()` to follow the convention of other method names.

mhrabovcin:
@akutz what is the reason for introducing session caching on EFS package level?

drivers/storage/efs/storage/efs_storage.go
akutz: Hi @mhrabovcin,

Nope. It returns the address of the instance ID's `ID` field or panics if the instance ID is not found in the context.

drivers/storage/efs/storage/efs_storage.go
akutz: Good idea.

akutz:
Hi @mhrabovcin,

You made two comments to which I'm unable to reply directly above, so please find my responses here after I include your original notes for convenience:

> I don't understand why is this driver adding API connection to EC2 service and not using it anywhere? If we want to reuse AWS service connection abstraction between different drivers I'd say it would be worth extracting this to a separate package and not have it in EFS driver.

So my original thought in adding the API connection was to use it to gather the default Security Group information in case it was omitted from the configuration file. However, after I originally saw your note and thought upon it I realized my initial idea was flawed as the place where the security group information is used makes the context not server-centric, but instance-centric. Therefore I intend to use raw HTTP GET commands via the executor and pass the default security group (if any) via the instance ID's `fields` property much like I now do with `region` and/or `availabilityZone` from the client/instance. I will be removing the API connection from the server code.

> what is the reason for introducing session caching on EFS package level?

The session caching is only secondary to the purpose of having client instance-scoped connections instead of one based on the server's configuration/instance information. Otherwise the server could never be outside the same zone as the clients which it manages. The caching was simply part of the parcel as it was something I had added on the EBS side for efficiency.

akutz:
Hi @mhrabovcin,

I retract my statement about removing the EC2 connection on the server-side. It's still required in order to map the security group name from the client to the ID required by the EFS `CreateMountTarget` operation. On the client side I can only grab the security group names via the special IP, not the SG IDs as is required by the EFS operation. Thus I need to do an `EC2.DescribeSecruityGroups` on the server side using creds in order to translate the security group name into its ID. All the more reason that the connection is based on client-side details.

akutz:
Hi @mhrabovcin,

I've validated that the [default security group](https://gist.github.com/akutz/9e24fc9ca3db072e6325206448003072) additions work as I hoped. If you don't have an strong objections, I'd like to merge this.

drivers/storage/efs/storage/efs_storage.go
mhrabovcin: Ok, i thought its a typo as its not being used anywhere in this patch.

mhrabovcin:
Hey @akutz, thanks for explaining the reason for including `ec2` API client. If I am reading the code correctly it means that if there is no security group specified in the driver configuration you'll try to guess the security group(s) from a instance.

I don't think this would work in usual way how security groups are structured. There is no reason to open NFS mounts inbound traffic on EC2 instances (and security groups applied to EC2 instance), which is the requirements for a MountTarget. In our experience usual approach is to create dedicated security group for a MountTarget(s) that will open only NFS port and allow to access a EFS instance from particular list of EC2 instances.

I'd suggest to error out if there is no security group specified instead of trying to guess (and probably guess incorrectly).

akutz:
Hi @mhrabovcin,

> I'd suggest to error out if there is no security group specified instead of trying to guess (and probably guess incorrectly).

Why do you think we're guessing the security group(s)? The names of the security groups to which an instance belongs are generated via the executor by curling `http://169.254.169.254/latest/meta-data/security-groups`, which returns a `\n` delimited list of the names of the security groups to which the requesting instance belongs. That data is then provided to the server via the instance ID's `Fields` property, which is then used in the `VolumeAttach` implementation [here](https://github.com/codedellemc/libstorage/blob/f54d639d3509860741aa5c95bc8be2a00ebc9e2f/drivers/storage/efs/storage/efs_storage.go#L610-L618). The names are [used](https://github.com/codedellemc/libstorage/blob/f54d639d3509860741aa5c95bc8be2a00ebc9e2f/drivers/storage/efs/storage/efs_storage.go#L640-L664) to get the security group IDs, which are then used in the EFS API call. 

I'm not sure that I understand where this is guessing. Are you referring to the fact that we're attempting to figure out which SG is intended as the export security group? I think that is probably what you mean here, right? If so, then yeah, I can see that. Perhaps we meet in the middle and try this logic:

1. If there is a security group named `efs` or `nfs` then that is the one assumed to be the default if no security group(s) is/are specified via the configuration file.
2. If there is not a security group named `efs` or `nfs` but there is a security group named `default`, then it is used if no security group(s) is/are specified via the configuration file.
3. If there is no security group named `efs`, `nfs`, or `default` and no security group(s) is/are specified in the configuration file on the server then an error is returned from the `VolumeAttach` operation.

Thoughts?

codenrhoden:
I think the idea of guessing comes from the fact that when you do get the list of security groups from `http://169.254.169.254/latest/meta-data/security-groups`, you still have to pick one to use. Which one?

So yeah, I agree with when you say:
>I'm not sure that I understand where this is guessing. Are you referring to the fact that we're attempting to figure out which SG is intended as the export security group? I think that is probably what you mean here, right? If so, then yeah, I can see that.

I'm not sure what the right amount of hand-holding here is. Looking for `efs`, `nfs`, `default`, in decreasing order of precedence, is still guessing. What if they all exist?

Unfortunately, I don't have the real-world experience with EFS to know what is standard/common here. Your heuristic could indeed be a middle ground, but if it is actually more likely to lead to confusion or using the wrong security group, it may indeed be better to require it be in the config file.

akutz:
How about this? If there is a single security group we use it, otherwise require the Config? Because if there is only one security group, there is no debate which one to use...

-- 
-a

codenrhoden:
@akutz: that makes sense to me!

akutz:
Hi @codenrhoden and @mhrabovcin,

I've updated the code so that the security group names are only used if there is exactly one of them. Additionally, I've introduced `efs.disableSessionCache` as a boolean config key that can be used to disable the session caching feature.

PR #345, opened 2016-11-21
Return volume from Docker integration Unmount()

codenrhoden:
This patch updates the IntegrationDriver types Unmount() method to
return a pointer to a Volume object. The unmount command is already
querying Libstorage to get an updated volume object after a detach has
occured, but it is not returning it.

Clients (like rexray) can use this returned value to display more up to
date status about the unmounted/detached volume without querying for it
a second time.

@akutz: this PR is a companion to https://github.com/codedellemc/rexray/pull/628. I fully expect you won't take it as-is since it makes a change to an existing Type. But I at least wanted to make the case for implementing something to this end, so that correct status can be shown in rexray.  If/when that happens, it may be better targeted for LS 0.4.x rather than 0.3.x.

### DIFF
--- a/api/registry/registry_integration.go
+++ b/api/registry/registry_integration.go
@@ -162,7 +162,7 @@ func (d *idm) Mount(
 func (d *idm) Unmount(
 	ctx types.Context,
 	volumeID, volumeName string,
-	opts types.Store) error {
+	opts types.Store) (*types.Volume, error) {
 
 	fields := log.Fields{
 		"volumeName": volumeName,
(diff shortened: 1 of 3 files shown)


akutz:
Hi @codenrhoden,

Let's discuss this when you're back. I'm getting a little concerned with how much we add to libStorage vs the clients. I like the fix, but I wonder if it shouldn't be in REX-Ray to request an updated status instead of having it buried in the integration driver.

akutz:
Hi @codenrhoden,

After reviewing the code more closely (which I should have done in the first place before commenting), this is a valuable and innocuous change. I like it. I'm going to do a little more testing first to make sure the signature change is in the right places. We have those driver manager shims in the middle, and I just want to make sure we caught those. Good addition Travis!

PR #347, opened 2016-11-29
Add Ceph RBD driver

codenrhoden:
This PR introduces the "rbd" storage driver, for interace with Ceph RBD devices.

This implementation does not use `go-ceph`, and therefore does not introduce new external dependencies. Instead it execs calls to the `rbd` executable, and parses the `json` output.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -588,6 +588,7 @@ remote storage systems. Currently the following storage drivers are supported:
 [VirtualBox](./storage-providers.md#virtualbox) | virtualbox
 [EBS](./storage-providers.md#aws-ebs) | ebs, ec2
 [EFS](./storage-providers.md#aws-efs) | efs
+[RBD](./storage-providers.md#ceph-rbd) | rbd
 ..more coming|
 
 The `libstorage.server.libstorage.storage.driver` property can be used to
@@ -694,6 +695,7 @@ ScaleIO|Yes
 VirtualBox|Yes
 EBS|Yes
 EFS|No
+RBD|No
 
 #### Ignore Used Count
 By default accounting takes place during operations that are performed
--- a/drivers/storage/rbd/tests/rbd_test.go
+++ b/drivers/storage/rbd/tests/rbd_test.go
@@ -1,0 +1,385 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_rbd
+
+package rbd
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+
+	log "github.com/Sirupsen/logrus"
+	gofig "github.com/akutz/gofig/types"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/codedellemc/libstorage/api/server"
+	apitests "github.com/codedellemc/libstorage/api/tests"
+	"github.com/codedellemc/libstorage/api/types"
+
+	// load the  driver
+	"github.com/codedellemc/libstorage/drivers/storage/rbd"
+	rbdx "github.com/codedellemc/libstorage/drivers/storage/rbd/executor"
+)
+
+const (
+	defaultPool = "rbd"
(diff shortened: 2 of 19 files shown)


codenrhoden:
So far I've been rebasing this PR and and keeping the whole thing in one commit. That makes it hard to see that I'm adding new things, though. So far now, I'm going to push new commits with major new additions so you can see where I'm at. 

I think that makes piecemeal reviews easier as well.

We can squash everything together at the end.

My latest commit adds tests in the `tests` folder.  The tests are in line with all the other drivers (with some simplification and enhancements for the "inspect" tests).  I think only two missing pieces now are docs, and a greater, scripted functional test that involves more than one server.

Code review is welcome!

akutz:
Hi @codenrhoden,

I don't rebase during the review process because the GitHub code review doesn't work if you make in-line comments and then update the branch with a rebased copy. I always perform additional commits during review -- at least until the very end.

Sounds good Travis!

.docs/user-guide/config.md
```diff
@@ -586,8 +586,9 @@ remote storage systems. Currently the following storage drivers are supported:
 [Isilon](./storage-providers.md#isilon) | isilon
 [ScaleIO](./storage-providers.md#scaleio) | scaleio
 [VirtualBox](./storage-providers.md#virtualbox) | virtualbox
-[EBS](./storage-providers.md#ebs) | ebs, ec2
-[EFS](./storage-providers.md#efs) | efs
+[EBS](./storage-providers.md#aws-ebs) | ebs, ec2
```
akutz, line 589: Good change, but please don't make it here as part of this PR. Please do it in a separate doc change.

.docs/user-guide/config.md
```diff
@@ -586,8 +586,9 @@ remote storage systems. Currently the following storage drivers are supported:
...
+[EFS](./storage-providers.md#aws-efs) | efs
```
akutz, line 590: Good change, but please don't make it here as part of this PR. Please do it in a separate doc change.

drivers/storage/rbd/tests/rbd_test.go
```diff
@@ -0,0 +1,370 @@
...
+
+func (a *dummyAddr) Network() string {
+	return "dummy"
+}
+
+func (a *dummyAddr) String() string {
+	return fmt.Sprintf("%s/24", a.ip.String())
+}
+
+var testIPs = []testIPInput{
```
akutz, line 111: Possibly optimize tests by parsing the unique IPs once and reusing the result.

drivers/storage/rbd/utils/utils.go
```diff
@@ -0,0 +1,287 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_rbd
```
akutz, line 1: Hi @codenrhoden,

I might add a sub-package called `rbd-utils` and make the file `rbd-utils.go` use the package main, thereby creating a new CLI. Then basically create a wrapper for this package so with the first arg of the CLI you can invoke a method from this file and the remaining args are the args for that function. 

I was just thinking it would be a quick and easy way to easily validate the functionality of the `utils` package.

codenrhoden:
I've removed the `WIP` designation on this PR, as I consider the work work to be feature-complete. I'm sure more things will be added, but it's ready for real review now. My latest commits added a `test-rbd` target to the `Makefile` for building the test binary, and also added a Vagrantfile and test plan in the `tests` directory.

codenrhoden:
@akutz.  I think we are all green here.  Take another look through when you can.

I did update the tests to use statically parsed IPs.

I can run my compiled tests package on the vagrant VM included in this PR as follows:

```
[vagrant@libstorage-rbd-test-admin ~]$ sudo ./rbd.test
PASS
coverage: 55.8% of statements in github.com/codedellemc/libstorage/drivers/storage/rbd, github.com/codedellemc/libstorage/drivers/storage/rbd/executor
```

PR #356, opened 2016-12-07
Build Tag Activated Drivers/Executors

akutz:
This patch enables the specific activation of storage drivers/executors via build tags. The tags `libstorage_storage_driver` and `libstorage_storage_executor`, if absent, will result in all drivers/executors being activated. This is the previous, default behavior. However, if the tags `libstorage_storage_driver` and `libstorage_storage_executor` are present along with `libstorage_storage_driver_NAME` and `libstorage_storage_executor_NAME`, where NAME is the name of a driver, ex. `ebs`, then only that driver and executor are considered for compilation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,13 +9,19 @@ go:
 os:
   - linux
 
+env:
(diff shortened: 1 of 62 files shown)


akutz:
Stupid Hound-CI. There are comments on each nil import! Ugh.

PR #372, opened 2016-12-26
Azure storage driver

Andrey-mp:
This patchset adds support of Azure Cloud.
There are several operations implemented in patchset:
create, attach, detach, remove (all base operations
for volumes).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -847,6 +847,7 @@ $1:
 	BUILD_TAGS="$$(BUILD_TAGS)" GOOS=$2 GOARCH=amd64 $$(MAKE) $$@
 $1-clean:
 	rm -f $1
+	rm -f $(EXECUTORS_GENERATED)
 GO_PHONY += $1-clean
 GO_CLEAN += $1-clean
 endif
--- a/drivers/storage/azure/azure.go
+++ b/drivers/storage/azure/azure.go
@@ -1,0 +1,103 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_azure
+
+package azure
+
+import (
+	gofigCore "github.com/akutz/gofig"
+	gofig "github.com/akutz/gofig/types"
+)
+
+const (
+	// Name is the provider's name.
+	Name = "azure"
+
+	// TagDelimiter separates tags from volume or snapshot names
+	TagDelimiter = "/"
+
+	// DefaultUseHTTPS - Use https prefix by default
+	// or not for Azure URI's
+	DefaultUseHTTPS = true
+
+	// TenantIDKey is a Directory ID from Azure
+	TenantIDKey = "tenantID"
+	// ClientIDKey is an Application ID from Azure
+	ClientIDKey = "clientID"
+	// ClientSecretKey is a secret of the application
+	ClientSecretKey = "clientSecret"
+	// CertPathKey is a path to application certificate in case of
+	// authorization via certificate
+	CertPathKey = "certPath"
+
+	// StorageAccountKey is a name of storage account
+	StorageAccountKey = "storageAccount"
+	// StorageAccessKey is an access key of storage account
+	StorageAccessKey = "storageAccessKey"
+	// TODO: add option to pass StorageURI
+
+	// SubscriptionIDKey is an ID of subscription
+	SubscriptionIDKey = "subscriptionID"
+	// ResourceGroupKey is a name of resource group
+	ResourceGroupKey = "resourceGroup"
+	// ContainerKey is a name of container in the storage account
+	// ('vhds' by default)
+	ContainerKey = "container"
+	// UseHTTPSKey is a flag about use https or not for making Azure URI's
+	UseHTTPSKey = "useHTTPS"
+	// TagKey is a tag key
+	TagKey = "tag"
+)
+
+const (
+	// ConfigAzure is a config key
+	ConfigAzure = Name
+
+	// ConfigAzureSubscriptionIDKey is a config key
+	ConfigAzureSubscriptionIDKey = ConfigAzure + "." + SubscriptionIDKey
+
+	// ConfigAzureResourceGroupKey is a config key
+	ConfigAzureResourceGroupKey = ConfigAzure + "." + ResourceGroupKey
+
+	// ConfigAzureTenantIDKey is a config key
+	ConfigAzureTenantIDKey = ConfigAzure + "." + TenantIDKey
+
+	// ConfigAzureStorageAccountKey is a config key
+	ConfigAzureStorageAccountKey = ConfigAzure + "." + StorageAccountKey
+
+	// ConfigAzureStorageAccessKeyKey is a config key
+	ConfigAzureStorageAccessKeyKey = ConfigAzure + "." + StorageAccessKey
+
+	// ConfigAzureContainerKey is a config key
+	ConfigAzureContainerKey = ConfigAzure + "." + ContainerKey
+
+	// ConfigAzureClientIDKey is a config key
+	ConfigAzureClientIDKey = ConfigAzure + "." + ClientIDKey
+
+	// ConfigAzureClientSecretKey is a config key
+	ConfigAzureClientSecretKey = ConfigAzure + "." + ClientSecretKey
+
+	// ConfigAzureCertPathKey is a config key
+	ConfigAzureCertPathKey = ConfigAzure + "." + CertPathKey
+
+	// ConfigAzureUseHTTPSKey is a config key
+	ConfigAzureUseHTTPSKey = ConfigAzure + "." + UseHTTPSKey
+
+	// ConfigAzureTagKey is a config key
+	ConfigAzureTagKey = ConfigAzure + "." + TagKey
+)
+
+func init() {
+	r := gofigCore.NewRegistration("Azure")
+	r.Key(gofig.String, "", "", "", ConfigAzureSubscriptionIDKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureResourceGroupKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureTenantIDKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureStorageAccountKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureContainerKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureClientIDKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureClientSecretKey)
+	r.Key(gofig.String, "", "", "", ConfigAzureCertPathKey)
+	r.Key(gofig.Bool, "", DefaultUseHTTPS, "", ConfigAzureUseHTTPSKey)
+	r.Key(gofig.String, "", "",
+		"Tag prefix for Azure naming", ConfigAzureTagKey)
+
+	gofigCore.Register(r)
+}
--- a/drivers/storage/azure/executor/azure_executor.go
+++ b/drivers/storage/azure/executor/azure_executor.go
@@ -70,7 +70,7 @@ var availLetters = []string{
 func (d *driver) NextDevice(
 	ctx types.Context,
 	opts types.Store) (string, error) {
-	// All possible device paths on Linux instances are /dev/sd[c-p]
+	// All possible device paths on Linux instances are /dev/sd[c-z]
 
 	// Find which letters are used for local devices
 	localDeviceNames := make(map[string]bool)
--- a/drivers/storage/azure/executor/azure_executor.go
+++ b/drivers/storage/azure/executor/azure_executor.go
@@ -1,0 +1,143 @@
+// +build !libstorage_storage_executor libstorage_storage_executor_azure
+
+package executor
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"path"
+	"regexp"
+	"strings"
+
+	gofig "github.com/akutz/gofig/types"
+	"github.com/akutz/goof"
+
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+	"github.com/codedellemc/libstorage/drivers/storage/azure"
+	"github.com/codedellemc/libstorage/drivers/storage/azure/utils"
+)
+
+// driver is the storage executor for the azure storage driver.
+type driver struct {
+	config gofig.Config
+}
+
+func init() {
+	registry.RegisterStorageExecutor(azure.Name, newDriver)
+}
+
+func newDriver() types.StorageExecutor {
+	return &driver{}
+}
+
+func (d *driver) Init(ctx types.Context, config gofig.Config) error {
+	ctx.Info("azure_executor: Init")
+	d.config = config
+	return nil
+}
+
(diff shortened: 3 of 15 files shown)


akutz:
Hi @Andrey-mp,

Do you mind rebasing this brand onto `master`? Thank you!

Makefile
```diff
@@ -847,6 +847,7 @@ $1:
 	BUILD_TAGS="$$(BUILD_TAGS)" GOOS=$2 GOARCH=amd64 $$(MAKE) $$@
 $1-clean:
 	rm -f $1
+	rm -f $(EXECUTORS_GENERATED)
```
akutz, line 850: Good catch!

drivers/storage/azure/azure.go
```diff
@@ -0,0 +1,99 @@
...
+	ContainerKey      = "container"
+	// UseHTTPS is a flag about use https or not for making Azure URI's
+	UseHTTPS          = "useHTTPS"
+	// TagKey is a tag key
+	TagKey            = "tag"
+)
+
+const (
+	// ConfigAZURE is a config key
+	ConfigAZURE                    = Name
```
akutz: Hi @Andrey-mp,

This is the only place I will mention this in my review, but why did you make `AZURE` capitalized throughout your use of it in constant/variable names? I checked, and it seems Microsoft doesn't use all-caps to write _Azure_, so I'm just wondering why you did?

image

drivers/storage/azure/executor/azure_executor.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+var errNoAvaiDevice = goof.New("no available device")
+
+// NextDevice returns the next available device.
+func (d *driver) NextDevice(
+	ctx types.Context,
+	opts types.Store) (string, error) {
+	// All possible device paths on Linux instances are /dev/sd[c-p]
+	letters := []string{
```
akutz: Since Go will keep malloc'd memory on the stack if it does not leak it probably makes more sense to keep this variable local to this function, but since mallocs are expensive in Go, I'm wondering if it makes sense to turn this into a package-scoped field that is allocated once on the heap.

drivers/storage/azure/executor/azure_executor.go
```diff
@@ -0,0 +1,141 @@
...
+
+	localDevices, err := d.LocalDevices(
+		ctx, &types.LocalDevicesOpts{Opts: opts})
+	if err != nil {
+		return "", goof.WithError("error getting local devices", err)
+	}
+	localDeviceMapping := localDevices.DeviceMap
+
+	for localDevice := range localDeviceMapping {
+		re, _ := regexp.Compile(`^/dev/` +
```
akutz: RegEx compilations are pretty expensive. The pieces of this regex are static, so why not use `regexp.MustCompile` once and assign the result to a package-level field. Compiling the regex in a loop seems pretty expensive.

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+	}
+
+	vm, err := d.getVM(ctx, vmName)
+	if err != nil {
+		return nil, goof.WithFieldsE(fields, "failed to detach volume to VM", err)
+	}
+
+	disks := *vm.StorageProfile.DataDisks
+	for i, disk := range disks {
+		if disk.Name != nil && *disk.Name == volumeID {
```
akutz: Hi @Andrey-mp,

I wasn't sure if you intended to do a case-sensitive comparison or not. If not then perhaps see [`strings.EqualFold(string, string) bool`](https://golang.org/pkg/strings/#example_EqualFold).

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+
+// Prefix volume or snapshot name with config tag
+func (d *driver) getFullName(name string) string {
+	if d.tag() != "" {
+		return d.tag() + azure.TagDelimiter + name
+	}
+	return name
+}
+
+// Retrieve config arguments
```
akutz, line 636: Hi @Andrey-mp,

I'm not going to make this comment on each of the below functions, but this remark does apply to all of the code below related to retrieving configuration properties.

There is no need to check the environment variables directly. Please see this [example](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/storage/ebs_storage.go#L1321-L1332) for the EBS driver (ignore the backwards-compatability checks). The [gofig](https://github.com/akutz/gofig) package uses @spf13's [Viper](https://github.com/spf13/viper), which in turn enables the assignment of environment variables to defined configuration properties. 

Let's look at [line \#87](https://github.com/cloudscaling/libstorage/blob/f4bf3809fe1e354dbccd7ef05086f06677c7ad6b/drivers/storage/azure/azure.go#L87) in your `azure.go`:

```go
r.Key(gofig.String, "", "", "", ConfigAZURESubscriptionIDKey)
```

In addition to notifying Gofig about the property `azure.subscriptionID`, the above function call instructs Viper to bind the property `azure.subscriptionID` to the CLI flag `--azureSubscriptionID` and the environment variable `AZURE_SUBSCRIPTIONID`. You can actually use Gofig to overwrite which CLI flag or environment variable the property is bound via the [`Key` function's trailing variadic parameter](https://github.com/akutz/gofig/blob/master/gofig_reg.go#L56-L114). 

For example, if we wanted to still use the config property `azure.subscriptionID` but change the CLI flag to `azureSubID` and the environment variable to `AZURE_SUB_ID` we could rewrite the above line like so:

```go
r.Key(gofig.String, "", "", "", 
	ConfigAZURESubscriptionIDKey,
	"azureSubID",
	"AZURE_SUB_ID")
```

Once the configuration properties are bound to a CLI flag or environment variable, invoking `config.GetString(azure.ConfigAZURESubscriptionIDKey)` checks first the environment variable, CLI flag argument, and finally a configuration property in that order.

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+			return nil, errGetLocDevs
+		}
+	}
+	*/
+
+	var volumesSD []*types.Volume
+	for _, blob := range *blobs {
+		volumeSD, error := d.toTypeVolume(ctx, &blob, attachments)
+		if error != nil {
+			ctx.WithError(error).Error("Failed to convert volume")
```
akutz: Hi @Andrey-mp,

Do you intend to have this error logged but not returned? If so perhaps log it as a warning instead?

akutz (changes requested):
Hi @Andrey-mp,

This is a great PR -- thank you so very much the obvious time and effort you spent on this. I have some suggestions/questions to which I hope you're able to respond below. 

One question I had regarding the entire PR is whether or not you used the GoMetalinter or at least go-fmt to ensure the files are consistent in their syntax/style? 

Also, if possible can you try and limit the width of lines to 80 chars. I may be pedantic, but I find it makes things easier to read.

Thank you again!

drivers/storage/azure/azure.go
```diff
@@ -0,0 +1,99 @@
...
+	// TODO: add option to pass StorageURI
+
+	// SubscriptionIDKey is an ID of subscription
+	SubscriptionIDKey = "subscriptionID"
+	// ResourceGroupKey is a name of resource group
+	ResourceGroupKey  = "resourceGroup"
+	// ContainerKey is a name of container in the storage account ('vhds' by default)
+	ContainerKey      = "container"
+	// UseHTTPS is a flag about use https or not for making Azure URI's
+	UseHTTPS          = "useHTTPS"
```
codenrhoden: Looks like both `UseHTTPS` and `StorageAccount` are used only when setting up config keys. To be consistent with all the other keys used in this manner, I feel like these should be renamed to `UseHTTPSKey` and `StorageAccountKey`.

drivers/storage/azure/executor/azure_executor.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		return fmt.Sprintf(
+			"/dev/%s%s", azureUtils.NextDeviceInfo.Prefix, letter), nil
+	}
+	return "", errNoAvaiDevice
+}
+
+const procPartitions = "/proc/partitions"
+
+var devRX = regexp.MustCompile(`^sd[a-z]$`)
```
codenrhoden, line 105: This is looking for `/dev/sd[a-z]`, but the comment in `NextDevice` says it can only be `/dev/sd[c-p]`. Are these both true?

If newly attached devices can only go to `p`, do we need to include all the way up to `z`? 

Functionally I'm sure this works, it just jumps out to me as an inconsistency.

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+		spt      *autorestAzure.ServicePrincipalToken
+		err      error
+	)
+
+	if d.tenantID == "" {
+		return nil, goof.New("Empty tenantID")
+	}
+
+	writeHkey(hkey, &d.subscriptionID)
+	writeHkey(hkey, &d.resourceGroup)
```
codenrhoden: resourceGroup doesn't seem to be a part of uniquely identifying a session. I don't think it needs to be a part of the hash key. It's never used in this function again.

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+
+// InstanceInspect returns an instance.
+func (d *driver) InstanceInspect(
+	ctx types.Context,
+	opts types.Store) (*types.Instance, error) {
+
+	iid := context.MustInstanceID(ctx)
+	return &types.Instance{
+		Name: iid.ID,
+		//Region:       iid.Fields[azure.InstanceIDFieldRegion],
```
codenrhoden: Seems like this can be removed for now?

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+	size *= size1GB
+
+	fields := map[string]interface{}{
+		"provider":      d.Name(),
+		"vmName":        vmName,
+		"volumeName":    volumeName,
+		"size_in_bytes": size,
+	}
+
+	volume, _ := d.getVolume(ctx, volumeName, types.VolAttNone)
```
codenrhoden: should an error here be ignored? What if the API failed and the check wasn't able to be performed?

drivers/storage/azure/storage/azure_storage.go
```diff
@@ -0,0 +1,914 @@
...
+		attachState = types.VolumeAttached
+	}
+	var attachmentsSD []*types.VolumeAttachment
+	if attachments.Requested() {
+		if attachState == types.VolumeAttached {
+			attachedInstanceID := types.InstanceID{
+				ID:     blob.Metadata["microsoftazurecompute_vmname"],
+				Driver: d.name,
+			}
+			if attachments.Mine() {
```
codenrhoden: Is it necessary for the Azure driver to do this filtering itself? I don't think that it is, and I don't think that it needs to handle setting `attachState` either. The framework handles this for you, here: https://github.com/codedellemc/libstorage/blob/release/0.4.0/api/server/router/volume/volume_routes.go#L131-L209

If a driver really does need to calculate `attachState` itself, it can do so. That will override the framework's default behavior of saying the volume is available if there are no attachments returned, it is attached if it is attached to the IID of the caller, or unavailable if attached to a different IID.

It is the intended behavior that `attachState` is not set if `attachments.Requested()` is false.

drivers/storage/azure/utils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+	if err := scanner.Err(); err != nil {
+		ctx.Debug("Specific file (/var/lib/dhcp/dhclient.eth0.leases) could not be read:")
+		ctx.Debug(err)
+	}
+	return false
+}
+
+// IsAzureInstance returns a flag indicating whether the executing host is an Azure
+// instance based on whether or not the metadata URL can be accessed.
```
codenrhoden: The comment here does not match how the function is actually implemented. It does not use metadata that I can see.

drivers/storage/azure/utils/utils.go
```diff
@@ -0,0 +1,77 @@
...
+			return nil, err
+		}
+		if !isAzure {
+			return nil, goof.New("Executing outside of Instance. InstanceID could not be obtained.")
+		}
+
+		// UUID can be obtained as descried in https://azure.microsoft.com/en-us/blog/accessing-and-using-azure-vm-unique-id/
+		// but this code will use hostname as ID
+
+		hostname, err = os.Hostname()
```
codenrhoden: How likely is this to happen? I'm guess that `AZURE_INSTANCE_ID` should normally be set. But falling back to a hostname seems iffy, unless Asure has requirements on whether hostnames have to be unique?

May be a non-issue, just curious.

codenrhoden (changes requested):
Echo'ing what @akutz said, thanks for the very thorough and complete PR! Hurray for tests.

Most of my comments are minor. My main feedback is that I don't think the filtering in toTypeVolume is necessary, and i also think that handling of the attachmentState can be handled by the framework instead of the driver.

drivers/storage/azure/azure.go
Andrey-mp: Will be renamed.

drivers/storage/azure/utils/utils.go
Andrey-mp: Azure doesn't set AZURE_INSTANCE_ID inside VM-s. This variable was implemented for unit tests only.
Azure set hostname inside of VM equals to name of VM in Azure engine. And this name is unique in the engine.
So because there is no other way to obtain instance identifier (that is suitable for REST API) then we get it from the hostname.

drivers/storage/azure/storage/azure_storage.go
Andrey-mp: Done.

drivers/storage/azure/executor/azure_executor.go
Andrey-mp: Done.

Andrey-mp:
Hi all.

We've thought that we fixed all your comments :)
Also I run gometalinter and it doesn't show me valuable errors.
We are waiting for next review.

drivers/storage/azure/executor/azure_executor.go
codenrhoden: This looks the same to me. `availLetters` is still defining [c-z], the regex is looking for [a-z], and the first comment inside of `NextDevice` says that it can be [c-p].

codenrhoden (changes requested):
Just have the one remaining comment on device names. I'll start playing with this on Azure!

I'd like to see a first attempt made at adding the driver into the docs, at .docs/user-guide/storage-providers.md.

Should be interesting, as I haven't used Azure before.

Andrey-mp:
Hi Travis,
We've got this code from the EBS driver that is very similar to Azure.
And in it code I see the same - NextDevice and LocalDevices has different letters...
And other executors have different behaviour for these functions also.

codenrhoden:
Closing in deference to #421

PR #374, opened 2017-01-03
rbd: require attach state to be available to attach

codenrhoden:
The `VolumeAttach` method was blindly proceeded with the steps to do an actual attach of the volume to the host.  I had incorrectly thought that `rexray` would not allow a call to attach a volume that wasn't available. Even if it did, it is incorrect to rely on `rexray` behavior to handle error cases -- `rexray` is just one of many possible clients, and it is up to the driver to enforce valid behavior.

This patch makes sure the volume is available before proceeding with the attach, and includes a testcase that catches the error. Multiple calls to `rbd map` of the same device on the same host will actually succeed, so we can test this scenario on a single host.

Before the fix:

```
[vagrant@libstorage-rbd-test-admin ~]$ sudo ./rbd.test -test.run VolumeAttach
--- FAIL: TestVolumeAttach (2.25s)
	tests.go:506: {
		  "attachmentState": 3,
		  "name": "6b8d1864",
		  "size": 8,
		  "id": "rbd.6b8d1864",
		  "type": "rbd"
		}
	tests.go:506: {
		  "attachments": [
		    {
		      "deviceName": "",
		      "instanceID": {
		        "id": "172.21.13.10",
		        "driver": "rbd"
		      },
		      "status": "",
		      "volumeID": "rbd.6b8d1864"
		    }
		  ],
		  "attachmentState": 2,
		  "name": "6b8d1864",
		  "size": 8,
		  "id": "rbd.6b8d1864",
		  "type": "rbd"
		}
	tests.go:506: {
		  "attachments": [
		    {
		      "deviceName": "",
		      "instanceID": {
		        "id": "172.21.13.10",
		        "driver": "rbd"
		      },
		      "status": "",
		      "volumeID": "rbd.6b8d1864"
		    }
		  ],
		  "attachmentState": 2,
		  "name": "6b8d1864",
		  "size": 8,
		  "id": "rbd.6b8d1864",
		  "type": "rbd"
		}
	tests.go:506: {
		  "attachments": [
		    {
		      "deviceName": "/dev/rbd0",
		      "instanceID": {
		        "id": "172.21.13.10",
		        "driver": "rbd"
		      },
		      "status": "",
		      "volumeID": "rbd.6b8d1864"
		    }
		  ],
		  "attachmentState": 2,
		  "name": "6b8d1864",
		  "size": 8,
		  "id": "rbd.6b8d1864",
		  "type": "rbd"
		}
	tests.go:506: {
		  "name": "6b8d1864",
		  "size": 8,
		  "id": "rbd.6b8d1864",
		  "type": "rbd"
		}
        Error Trace:    rbd_test.go:395
			rbd_test.go:416
			tests.go:290
			asm_amd64.s:2086
	Error:		An error is expected but got nil.

	rbd_test.go:397: volumeAttach succeeded when it should have failed
FAIL
coverage: 6.5% of statements in github.com/codedellemc/libstorage/drivers/storage/rbd, github.com/codedellemc/libstorage/drivers/storage/rbd/executor
```

After the fix:

```
[vagrant@libstorage-rbd-test-admin ~]$ sudo ./rbd.test -test.run VolumeAttach
PASS
coverage: 6.5% of statements in github.com/codedellemc/libstorage/drivers/storage/rbd, github.com/codedellemc/libstorage/drivers/storage/rbd/executor
```

### DIFF
--- a/drivers/storage/rbd/storage/rbd_storage.go
+++ b/drivers/storage/rbd/storage/rbd_storage.go
@@ -252,12 +252,27 @@ func (d *driver) VolumeAttach(
 		return nil, "", goof.WithError("Unable to set image name", err)
 	}
 
+	vol, err := d.VolumeInspect(
+		ctx, volumeID, &types.VolumeInspectOpts{
+			Attachments: types.VolAttReq,
+		})
+	if err != nil {
+		return nil, "", goof.WithError("error getting volume", err)
+	}
+
+	if vol.AttachmentState != types.VolumeAvailable {
+		if !opts.Force {
+			return nil, "",
+				goof.New("volume in wrong state for attach")
+		}
+	}
+
 	_, err = utils.RBDMap(pool, imageName)
 	if err != nil {
 		return nil, "", err
 	}
 
-	vol, err := d.VolumeInspect(ctx, volumeID,
+	vol, err = d.VolumeInspect(ctx, volumeID,
 		&types.VolumeInspectOpts{
 			Attachments: types.VolAttReqTrue,
 		},
--- a/drivers/storage/rbd/tests/rbd_test.go
+++ b/drivers/storage/rbd/tests/rbd_test.go
@@ -385,6 +385,23 @@ func volumeInspectAvailable(
 	return reply
 }
 
+func volumeAttachFail(
+	t *testing.T, client types.Client, volumeID string) *types.Volume {
+
+	log.WithField("volumeID", volumeID).Info("attaching volume")
+	reply, _, err := client.API().VolumeAttach(
+		nil, rbd.Name, volumeID, &types.VolumeAttachRequest{})
+
+	assert.Error(t, err)
+	if err == nil {
+		t.Error("volumeAttach succeeded when it should have failed")
+		t.FailNow()
+	}
+	apitests.LogAsJSON(reply, t)
+
+	return reply
+}
+
 func TestVolumeAttach(t *testing.T) {
 	if skipTests() {
 		t.SkipNow()
@@ -396,6 +413,7 @@ func TestVolumeAttach(t *testing.T) {
 		_ = volumeInspectAttached(t, client, vol.ID)
 		_ = volumeInspectAttachedDevices(t, client, vol.ID)
 		_ = volumeInspectNoAttachments(t, client, vol.ID)
+		_ = volumeAttachFail(t, client, vol.ID)
 		_ = volumeDetach(t, client, vol.ID)
 		_ = volumeInspectDetached(t, client, vol.ID)
 		_ = volumeInspectAvailable(t, client, vol.ID)


akutz:
Hi @codenrhoden,

Let's take a moment and review the assumptions that were made. We can possibly enhance libStorage/REX-Ray based on them:

## `VolumeAttach` & Unavailable Volumes
You noted surprise at the `VolumeAttach` function willingness to act on a volume in an invalid state for an attach operation. Originally I had designed libStorage to query all volumes prior to any operation on them. That way it would be possible to validate state or other requirements prior to an operation on an object. However, I decided that it was up to a client to perform such checks as querying a volume could be an expensive operation, wholly dependent on the underlying storage platform. 

Thus, as you discovered, if a client invokes the API for attaching a volume, all the client is sending is the volume's ID. The server-side then provides that ID to the storage driver. Up to that point, the server's framework has not queried the volume ID at all to validate that it exists or is in a state to be acted upon by the intended operation. 

That's why the Attach API call, and others, have the [possible HTTP response codes](http://docs.libstorage.apiary.io/#reference/volumes/volume-inspector/attach) that include 404 -- resource not found. 

I believe that this is the correct and desired behavior. Otherwise a blanket check could prove to be incredibly expensive. Thoughts?

## The Patch
Is there a reason that this patch is required? Why is it wrong for the attach operation simply to fail if the volume is in an invalid state? 

## REX-Ray's Behavior
Per @cduchesne on behalf of you and the RBD work you're performing, I had updated the REX-Ray CLI to request already-attached volumes when performing the `volume attach` command for purposes of making the `attach`, `detach`, `mount, and `unmount` operations  idempotent.  This work was noted in REX-Ray PRs [\#651](https://github.com/codedellemc/rexray/pull/651) and [\#658](https://github.com/codedellemc/rexray/pull/658).

However, even if the idempotent flag is set to true, the code in the REX-Ray repository [at lines 313-318](https://github.com/codedellemc/rexray/blob/master/cli/cli/cmds_10_volume.go#L313-L318) should simply emit the the volume without performing an operation on it.

Are you testing with RR 0.6.3, `master`, a 0.6.4 RC, or a 0.7.0 RC?

codenrhoden:
Hi @akutz,

>I believe that this is the correct and desired behavior. Otherwise a blanket check could prove to be incredibly expensive. Thoughts?

yes, I think this is correct.  I wasn't surprised by libStorage behavior, rather, REX-Ray's.  @cduchesne provided we with logs from these commands that showed that the RBD driver was happy to attach volumes that were listed as unavailabe:

```
[root@rexray-rbd1 ~]# rexray volume ls
ID           Name     Status       Size
rbd.redis01  redis01  unavailable  8
rbd.vol01    vol01    attached     5
[root@rexray-rbd1 ~]# rexray volume mount redis01
ID           Name     Status    Size  Path
rbd.redis01  redis01  attached  8     /var/lib/libstorage/volumes/redis01/data
```
I was surprised that rexray made the call to attach a volume in this case. But regardless, the driver should reject it, and that is what this patch adds.

> Is there a reason that this patch is required? Why is it wrong for the attach operation simply to fail if the volume is in an invalid state?

There is nothing wrong with that operation to fail -- it should. But it wasn't failing, as you can see from the paste above. So that's what this patch adds. It just brings the logic in line with drivers like ScaleIO: https://github.com/codedellemc/libstorage/blob/master/drivers/storage/scaleio/storage/scaleio_storage.go#L482-L499

There are a few reasons why I didn't do this to begin with, but they don't really matter. :)

> Are you testing with RR 0.6.3, master, a 0.6.4 RC, or a 0.7.0 RC?

I'm not sure what version @cduchesne was using, but since I was able to recreate the issue without rexray (just using testing) that's what I did. The client was unimportant.

Hopefully this addresses your questions.

akutz:
Hi @codenrhoden,

Well this shouldn't be occurring. Period. Not if @cduchesne is using the latest RR bits. This is disconcerting. @cduchesne, I thought this is what we had tested and fixed with PR #369 and libStorage 0.3.7. @codenrhoden, please take a look at PR #369 as it should be exactly what you are seeing and the fix for it. Both RR releases [0.7.0-rc3](https://github.com/codedellemc/rexray/releases/tag/v0.7.0-rc3) and [0.6.4-rc4](https://github.com/codedellemc/rexray/releases/tag/v0.6.4-rc4) should include the libStorage commit 795b043065f39bf59fa684d3f7d2fb0387c27db1 which is the fix for the above issue.

codenrhoden:
@akutz: okay, will do.

That bit aside, do you think that drivers should protect against this (say, a client that isn't rexray). I think the PR is still valid. Do you?

akutz:
Hi @codenrhoden,

I think it's up to the driver. I see no real reason to shield against it as it's an extra call that may not actually protect against anything since there is no locking. If you want to do it, go for it. It probably actually does provide some value on a single-system implementation such as RBD.

codenrhoden:
@akutz  Okay, I'd still like it to go in if it can, then.

As for testing with the latest RC, you are correct that rexray no longer allows this:

```
[root@libstorage-rbd-test-admin ~]# rexray volume ls
ID            Name      Status       Size
rbd.test      test      unavailable  1
[root@libstorage-rbd-test-admin ~]# rexray volume attach test
FATA[0002] no volumes found
[root@libstorage-rbd-test-admin ~]# rexray version
REX-Ray
-------
Binary: /usr/bin/rexray
Flavor: client+agent+controller
SemVer: 0.7.0-rc3
OsArch: Linux-x86_64
Branch: v0.7.0-rc3
Commit: adba1c1b84895a3d4376394e620f75ef376de26b
Formed: Wed, 21 Dec 2016 23:29:05 UTC

libStorage
----------
SemVer: 0.4.0-rc3
OsArch: Linux-x86_64
Branch: v0.7.0-rc3
Commit: 04404c8b0e83d3eee9119a63291f92b914963a54
Formed: Wed, 21 Dec 2016 23:28:48 UTC
```

PR #384, opened 2017-01-05
Add storage driver template

codenrhoden:
Introduce a storage driver template to make it easier for developers to
start adding a new driver from scratch. Includes all required public
methods except for the `Login` method in the storage package.

### DIFF
--- a/drivers/storage/template/tests/DRIVERNAME_test.go
+++ b/drivers/storage/template/tests/DRIVERNAME_test.go
@@ -1,0 +1,44 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_DRIVERNAME
+
+package DRIVERNAME
+
+import (
+	"os"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/codedellemc/libstorage/api/server"
+	"github.com/codedellemc/libstorage/api/types"
+)
+
+// Put contents of sample config.yml here
+var (
+	configYAML = []byte(``)
+)
+
+var volumeName string
+var volumeName2 string
+
+// Check environment vars to see whether or not to run this test
+func skipTests() bool {
+	travis, _ := strconv.ParseBool(os.Getenv("TRAVIS"))
+	noTest, _ := strconv.ParseBool(os.Getenv("TEST_SKIP_DRIVERNAME"))
+	return travis || noTest
+}
+
+// Set volume names to first part of UUID before the -
+func init() {
+	uuid, _ := types.NewUUID()
+	uuids := strings.Split(uuid.String(), "-")
+	volumeName = uuids[0]
+	uuid, _ = types.NewUUID()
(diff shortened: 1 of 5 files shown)


codenrhoden:
@akutz -- what do you think, useful?

Trying to make it easier to create the scaffold of a driver from scratch.

drivers/storage/template/tests/DRIVERNAME_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/codedellemc/libstorage/api/server"
+	"github.com/codedellemc/libstorage/api/types"
+)
+
+// Put contents of sample config.yml here
+var (
+	configYAML = []byte(``)
+)
+
+var volumeName string
```
akutz, line 20: I'd remove the volumeName and volumeName2 as they're just going to confuse people.

> Are we supposed to use those? What are those for?

People can be very literal.

akutz:
Hi @codenrhoden,

First off, my feedback is not an attempt to rain on our parade or diminish the effort, time, or intent of what you've done here. All that is great. That said...

I think the value you're intending this provide would likely be better communicated via documentation. Plenty of people have written drivers based on reviewing existing ones or the VFS driver while plenty other people have had issues translating the existing driver code into instructions on how to create a new driver.

My thoughts are that while this is neat, I don't think it really helps someone know how to write a new driver more than telling them to look at existing drivers. Whereas developer docs explaining how to create a driver and the purpose of each function/pkg/etc would be the most useful.

What do you think?

codenrhoden:
@akutz 

I absolutely agree that the most impactful way to teach someone how to create a new driver is through documentation. It was not my intent for this to stand in for that.

My intent with this is to make it easier to get started with actual code. When I start a new driver from scratch I copy an entire driver tree from an existing driver, then delete all code to leave me with the required functions to implement the interface.

I had just done that as I am starting a new driver again, and was struck that it took me a few minutes, and it might be worthwhile to not have to do it again. I do also think it is nice to look at the "empty" files to get a feel for the scope of the work. You can see all the functions you need to implement, and see what they are for.

So this was really meant to be more of a convenient time saver, not a teaching tool. I will likely keep this branch locally for my personal use no matter what. Totally up to you if you think it provides value to other/new developers.

No worries about hurting my feelings or diminishing anything. It was ~10 minutes of work, 9 of which were things I was doing anyways. :)

akutz:
So how about this then @codenrhoden, let's add a README.md to the root of this template with instructions or a link to the documentation you/we/I create in the RTFD site that shows how to use the template to create a new driver, and the purpose of the diff pkgs, funcs, etc?

codenrhoden:
@akutz That sounds like a great plan, and I will add that!

codenrhoden:
superseded by https://github.com/codedellemc/libstorage/pull/550

PR #390, opened 2017-01-17
Renamed Docker Int Driver to Linux Int Driver

akutz:
The Docker Integration driver has been renamed to the more appropriate Linux Integration driver. All of the functionality is the same except the driver's key is now `linux` instead of `docker` and the package is `github.com/codedellemc/libstorage/drivers/integration/linux`.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -610,9 +610,9 @@ integration drivers are supported:
 
  Driver | Driver Name
 --------|------------
-Docker   | docker
+Linux   | linux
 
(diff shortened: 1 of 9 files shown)


codenrhoden (approved):
I think renaming this makes it's purpose much more clear. I know there is a bit of a question on what to rename it to. I for one think `linux` is just fine. This driver covers how you deal with volumes on linux machines. You could certainly call it a `reference` integration driver, but it's still only a reference implementation for Linux (IMO). I wouldn't call it `standard` as that implies to me more of a standards definition to be implemented. My two cents.

PR #392, opened 2017-01-18
DigitalOcean driver

protochron:
This implements a first-pass at a DigitalOcean storage provider. Right now it supports everything except for snapshots which I'm planning on adding as a followup PR.

TODOs:
* [x] fix the tests
* [x] update the README

### DIFF
--- a/drivers/storage/digitalocean/digitalocean.go
+++ b/drivers/storage/digitalocean/digitalocean.go
@@ -1,0 +1,47 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_digitalocean
+
+package digitalocean
+
+import (
+	gofigCore "github.com/akutz/gofig"
+	gofig "github.com/akutz/gofig/types"
+)
+
+const (
+	// Name is the name of the driver
+	Name = "digitalocean"
+
+	// InstanceIDFieldRegion is the key used to retrive the region from the
+	// instance id map
+	InstanceIDFieldRegion = "region"
+
+	// InstanceIDFieldName is the key used to retrive the name from the instance
+	// id map
+	InstanceIDFieldName = "name"
+
+	// VolumePrefix is the value that every DO volume appears with DigitalOcean
+	// volumes are are found using disk/by-id, ex:
+	// /dev/disk/by-id/scsi-0DO_Volume_volume-nyc1-01 See
+	// https://www.digitalocean.com/community/tutorials/how-to-use-block-storage-on-digitalocean#preparing-volumes-for-use-in-linux
+	VolumePrefix = "scsi-0DO_Volume_"
+
+	// ConfigDOToken is the key for the token in the config file
+	ConfigDOToken = Name + ".token"
+
+	// ConfigDORegion is the key for the region in the config file
+	ConfigDORegion = Name + ".region"
+)
+
+func init() {
+	registerConfig()
+}
+
+func registerConfig() {
+	r := gofigCore.NewRegistration("DigitalOcean")
+	r.Key(gofig.String, "", "", "",
+		ConfigDOToken,
+		"digitaloceanAccessToken",
+		"DIGITALOCEAN_ACCESS_TOKEN")
+	r.Key(gofig.String, "", "", "DigitalOcean region", ConfigDORegion)
+	gofigCore.Register(r)
+}
--- a/drivers/storage/digitalocean/executor/digitalocean_executor.go
+++ b/drivers/storage/digitalocean/executor/digitalocean_executor.go
@@ -1,0 +1,92 @@
+// +build !libstorage_storage_executor libstorage_storage_executor_digitalocean
+
+package executor
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"regexp"
+
+	gofig "github.com/akutz/gofig/types"
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+	do "github.com/codedellemc/libstorage/drivers/storage/digitalocean"
+	doUtils "github.com/codedellemc/libstorage/drivers/storage/digitalocean/utils"
+)
+
+var (
+	diskPrefix = regexp.MustCompile(`^` + do.VolumePrefix + `(.*)`)
+	diskSuffix = regexp.MustCompile("part-.*$")
+)
+
+type driver struct {
+	config gofig.Config
+}
+
+func init() {
+	registry.RegisterStorageExecutor(do.Name, newDriver)
+}
+
+func newDriver() types.StorageExecutor {
+	return &driver{}
+}
+
+func (d *driver) Name() string {
+	return do.Name
+}
+
+func (d *driver) Init(ctx types.Context, config gofig.Config) error {
+	d.config = config
+	return nil
+}
+
+func (d *driver) InstanceID(
+	ctx types.Context, opts types.Store) (*types.InstanceID, error) {
+	return doUtils.InstanceID(ctx)
+}
+
+func (d *driver) NextDevice(
+	ctx types.Context, opts types.Store) (string, error) {
+	return "", types.ErrNotImplemented
+}
+
+func (d *driver) LocalDevices(
+	ctx types.Context,
+	opts *types.LocalDevicesOpts) (*types.LocalDevices, error) {
+	deviceMap := map[string]string{}
+	diskIDPath := "/dev/disk/by-id"
+
+	dir, _ := ioutil.ReadDir(diskIDPath)
+	for _, device := range dir {
+		switch {
+		case !diskPrefix.MatchString(device.Name()):
+			continue
+		case diskSuffix.MatchString(device.Name()):
+			continue
+		case diskPrefix.MatchString(device.Name()):
+			volumeName := diskPrefix.FindStringSubmatch(device.Name())[1]
+			devPath, err := filepath.EvalSymlinks(
+				fmt.Sprintf("%s/%s", diskIDPath, device.Name()))
+			if err != nil {
+				return nil, err
+			}
+			deviceMap[volumeName] = devPath
+		}
+	}
+
+	ld := &types.LocalDevices{Driver: d.Name()}
+	if len(deviceMap) > 0 {
+		ld.DeviceMap = deviceMap
+	}
+
+	return ld, nil
+}
+
+func (d *driver) Supported(ctx types.Context, opts types.Store) (bool, error) {
+	return doUtils.IsDroplet(ctx)
+}
+
+func (d *driver) token() string {
+	return d.config.GetString(do.ConfigDOToken)
+}
--- a/drivers/storage/digitalocean/storage/digitalocean_storage.go
+++ b/drivers/storage/digitalocean/storage/digitalocean_storage.go
@@ -1,0 +1,362 @@
+// +build !libstorage_storage_driver libstorage_storage_driver_digitalocean
+
+package storage
+
+import (
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	gofig "github.com/akutz/gofig/types"
+	"github.com/akutz/goof"
+
+	"github.com/digitalocean/godo"
+
+	"github.com/codedellemc/libstorage/api/context"
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+
+	do "github.com/codedellemc/libstorage/drivers/storage/digitalocean"
+	doUtils "github.com/codedellemc/libstorage/drivers/storage/digitalocean/utils"
+)
+
+type driver struct {
+	name   string
+	config gofig.Config
+	client *godo.Client
+}
+
+func init() {
+	registry.RegisterStorageDriver(do.Name, newDriver)
+}
+
+func newDriver() types.StorageDriver {
+	return &driver{name: do.Name}
+}
+
+func (d *driver) Name() string {
+	return do.Name
+}
+
+func (d *driver) Init(ctx types.Context, config gofig.Config) error {
+	d.config = config
+	token := d.config.GetString(do.ConfigDOToken)
+
+	fields := log.Fields{
+		"token": token,
+	}
+
+	if token == "" {
+		fields["token"] = ""
+	} else {
+		fields["token"] = "******"
+	}
+
+	fields["region"] = d.config.GetString(do.ConfigDORegion)
+
+	client, err := doUtils.Client(token)
+	if err != nil {
+		return err
+	}
+	d.client = client
+
+	log.WithFields(fields).Info("storage driver initialized")
+
+	return nil
+}
+
+func (d *driver) Type(ctx types.Context) (types.StorageType, error) {
+	return types.Block, nil
+}
+
+// DigitalOcean volumes are are found using device-by-id, ex:
+// /dev/disk/by-id/scsi-0DO_Volume_volume-nyc1-01 See
+// https://www.digitalocean.com/community/tutorials/how-to-use-block-storage-on-digitalocean#preparing-volumes-for-use-in-linux
+func (d *driver) NextDeviceInfo(
+	ctx types.Context) (*types.NextDeviceInfo, error) {
+	return nil, nil
+}
+
+func (d *driver) InstanceInspect(
+	ctx types.Context, opts types.Store) (*types.Instance, error) {
+	iid := context.MustInstanceID(ctx)
+	return &types.Instance{
+		InstanceID:   iid,
+		Region:       iid.Fields[do.InstanceIDFieldRegion],
+		Name:         iid.Fields[do.InstanceIDFieldName],
+		ProviderName: iid.Driver,
+	}, nil
+}
+
+func (d *driver) Volumes(
+	ctx types.Context, opts *types.VolumesOpts) ([]*types.Volume, error) {
+	doVolumes, _, err := d.client.Storage.ListVolumes(nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var volumes []*types.Volume
+	for _, vol := range doVolumes {
+		volumes = append(volumes, d.toTypesVolume(ctx, &vol, opts.Attachments))
+	}
+
+	return volumes, nil
+}
+
+func (d *driver) VolumeInspect(
+	ctx types.Context, volumeID string, opts *types.VolumeInspectOpts) (*types.Volume, error) {
+	doVolume, _, err := d.client.Storage.GetVolume(volumeID)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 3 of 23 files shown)


drivers/storage/digitalocean/executor/digitalocean_executor.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+		}
+	}
+
+	return false, nil
+}
+
+func (d *driver) token() string {
+	var token string
+	envToken := os.Getenv("DIGITALOCEAN_ACCESS_TOKEN")
```
akutz: Hi @protochron,

I'm not going to make this comment on each of the below calls, but this remark does apply to all of the code below related to retrieving configuration properties.

There is no need to check the environment variables directly. Please see this [example](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/storage/ebs_storage.go#L1321-L1332) for the EBS driver (ignore the backwards-compatability checks). The [gofig](https://github.com/akutz/gofig) package uses @spf13's [Viper](https://github.com/spf13/viper), which in turn enables the assignment of environment variables to defined configuration properties. 

Let's look at [line \#32](https://github.com/protochron/libstorage/blob/f7ce053cd9fec2638ca55cefa79abacfffd83495/drivers/storage/digitalocean/digitalocean.go#L32) in your `digitalocean.go`:

```go
r.Key(gofig.String, "", "", "", Name+"."+"token")
```

In addition to notifying Gofig about the property `digitalocean.token`, the above function call instructs Viper to bind the property `digitalocean.token` to the CLI flag `--digitaloceanToken` and the environment variable `DIGITALOCEAN_TOKEN`. You can actually use Gofig to overwrite which CLI flag or environment variable the property is bound via the [`Key` function's trailing variadic parameter](https://github.com/akutz/gofig/blob/master/gofig_reg.go#L56-L114). 

For example, if we wanted to still use the config property `digitalocean.token` but change the CLI flag to `digiToken` and the environment variable to `DIGI_TOKEN` we could rewrite the above line like so:

```go
r.Key(gofig.String, "", "", "", 
	Name+"."+"token",
	"digiToken",
	"DIGI_TOKEN")
```

Once the configuration properties are bound to a CLI flag or environment variable, invoking `config.GetString("digitalocean.token")` checks first the environment variable, CLI flag argument, and finally a configuration property in that order.

drivers/storage/digitalocean/storage/digitalocean_storage.go
```diff
@@ -0,0 +1,319 @@
...
+
+func mustInstanceIDID(ctx types.Context) *string {
+	return &context.MustInstanceID(ctx).ID
+}
+
+func (d *driver) toTypesVolume(ctx types.Context, volume *godo.Volume, attachments types.VolumeAttachmentsTypes) *types.Volume {
+	// Collect attachment info for the volume
+	var atts []*types.VolumeAttachment
+	if attachments.Requested() {
+		for _, id := range volume.DropletIDs {
```
protochron: This is the only good way you can find out whether or not a volume is attached. If you're using `VolumeInspect` with `VolAttReqTrue` like we do in the tests, this causes it to skip adding an attachments block entirely which doesn't seem to be expected behavior. Should it default to adding a block anyway even if the volume isn't attached to a host?

drivers/storage/digitalocean/storage/digitalocean_storage.go
```diff
@@ -0,0 +1,319 @@
...
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/digitalocean/godo"
+
+	gofig "github.com/akutz/gofig/types"
+	"github.com/akutz/goof"
+	"github.com/codedellemc/libstorage/api/context"
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+	"github.com/codedellemc/libstorage/drivers/storage/digitalocean"
```
akutz: Hi @protochron,

Maybe import this as a shorter alias?

```go
digocn "github.com/codedellemc/libstorage/drivers/storage/digitalocean"
```

That way it reduces line width where the package is referenced?

drivers/storage/digitalocean/storage/digitalocean_storage.go
```diff
@@ -0,0 +1,319 @@
...
+	"github.com/codedellemc/libstorage/drivers/storage/digitalocean/utils"
```
akutz: Hi @protochron,

Maybe import this as a non-generic utils alias?

```go
doUtils "github.com/codedellemc/libstorage/drivers/storage/digitalocean/utils"
```

That way it is clear what utils package is used when referencing it in the file.

drivers/storage/digitalocean/tests/digitalocean_test.go
```diff
@@ -0,0 +1,365 @@
...
+	log.WithField("volumeID", volumeID).Info("inspecting volume")
+	reply, err := client.API().VolumeInspect(nil, digitalocean.Name, volumeID, 0)
+	assert.NoError(t, err)
+
+	if err != nil {
+		t.Error("failed volumeInspectDetachedFail")
+		t.FailNow()
+	}
+	apitests.LogAsJSON(reply, t)
+	assert.Len(t, reply.Attachments, 1)
```
protochron: This currently fails since `VolumeInspect` is not returning attachments if the volume isn't attached anywhere. See the comment I left in the `VolumeInspect` implementation for details.

akutz (commented):
Hi @protochron,

This is a great PR -- thank you so very much the obvious time and effort you spent on this. I have some suggestions/questions to which I hope you're able to respond below.

One question I had regarding the entire PR is whether or not you used the GoMetalinter or at least go-fmt to ensure the files are consistent in their syntax/style?

Also, if possible can you try and limit the width of lines to 80 chars. I may be pedantic, but I find it makes things easier to read.

Thank you again!

drivers/storage/digitalocean/storage/digitalocean_storage.go
akutz: Hi @protochron,

Ah, I see the issue. Hang on.

akutz: Hi @protochron,

So the issue is that your tests are perhaps using an over-eager attachments mask, thus causing volumes to be filtered out by the framework and not returned to you. This is by design. See the [attachments related querying tests](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/vfs/tests/vfs_test.go#L151-L377) for the VFS storage driver. I use the VFS driver as the default means of validating the framework since VFS is pure-software with no external dependencies.

What you need to do is review the bit mask constants for querying volumes with attachments. 

* [`VolumeAttachmentTypes` code definitions](https://github.com/codedellemc/libstorage/blob/master/api/types/types_drivers_storage.go#L11-L38)
* libStorage API reference for [querying volumes with attachment information](http://docs.libstorage.apiary.io/#reference/volumes/volumes-collection/get-with-attachments)

Finally, here is a series of [predefined masks](https://github.com/codedellemc/libstorage/blob/master/api/types/types_drivers_storage.go#L40-L125) for the attachment bits. Notice that `VolAttReqTrue` is really an alias for `VolAttReqWithDevMapOnlyVolsAttachedToInstance` which is `1 | 2 | 4 | 8` and requests attachment information before returning only volumes attached to the instance provided via the InstanceID header with an attempt to match the attachments to the provided local device information via the local devices header.

protochron:
@akutz I went through and wrapped everything to 80 characters. I write everything using `vim-go` which runs go-fmt on every save. Did something stand out to you from a stylistic standpoint?

drivers/storage/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,47 @@
...
+package digitalocean
+
+import (
+	gofigCore "github.com/akutz/gofig"
+	gofig "github.com/akutz/gofig/types"
+)
+
+const (
+	// Name is the name of the driver
+	Name = "digitalocean"
```
akutz, line 12: Hi @protochron,

Does DigitalOcean have any type of abbreviation you'd like to use here? To make configuration properties a little easier/shorter to read/write?

drivers/storage/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,47 @@
...
+func init() {
+	registerConfig()
+}
+
+func registerConfig() {
+	r := gofigCore.NewRegistration("DigitalOcean")
+	r.Key(gofig.String, "", "", "",
+		ConfigDOToken,
+		"digitaloceanAccessToken",
+		"DIGITALOCEAN_ACCESS_TOKEN")
```
akutz, line 44: Hi @protochron,

Is there a reason you explicitly specified the CLI flag and environment variable names here?

drivers/storage/digitalocean/executor/digitalocean_executor.go
```diff
@@ -0,0 +1,120 @@
...
+	gofig "github.com/akutz/gofig/types"
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+	do "github.com/codedellemc/libstorage/drivers/storage/digitalocean"
+	doUtils "github.com/codedellemc/libstorage/drivers/storage/digitalocean/utils"
+)
+
+var (
+	diskPrefix = regexp.MustCompile(
+		fmt.Sprintf("^%s(.*)", do.VolumePrefix))
```
akutz: Don't use `fmt.Sprintf` here as `do.VolumePrefix` is a constant. The code as it is malloc's two new strings just to create a regex. Since `do.VolumePrefix` is a constant do this:

```go
diskPrefix = regexp.MustCompile(`^` + do.VolumePrefix + `(.*)`)
```

Golang lacks native string interning, so I avoid creating new ones where possible when you're just going to throw them away as an initialization piece for some other object.

drivers/storage/digitalocean/executor/digitalocean_executor.go
```diff
@@ -0,0 +1,120 @@
...
+	ctx types.Context, opts types.Store) (*types.InstanceID, error) {
+	return doUtils.InstanceID(ctx)
+}
+
+func (d *driver) NextDevice(
+	ctx types.Context, opts types.Store) (string, error) {
+	return "", types.ErrNotImplemented
+}
+
+func (d *driver) LocalDevices(
```
akutz, line 54: Hi @protochron,

While none of the drivers make it a point to note this, and some don't do it, the `LocalDevices` function is meant only to return devices relevant to the relevant storage platform.

drivers/storage/digitalocean/storage/digitalocean_storage.go
```diff
@@ -0,0 +1,336 @@
...
+import (
+	"fmt"
+	"path/filepath"
+	"strconv"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/digitalocean/godo"
+
+	gofig "github.com/akutz/gofig/types"
```
akutz: Hi @protochron,

I won't repeat this comment in other files, but it's relevant to them. I prefer import blocks to be like so:

```go
import (
	stdlib

	external_pkgs

	external_pkgs_relevant_to_storage_driver

	project_pkgs

	project_pkgs_relevant_to_storage_driver
```

For this file's imports I'd organize them like so:

```go
import (
	"fmt"
	"path/filepath"
	"strconv"
	"time"

	gofig "github.com/akutz/gofig/types"
	"github.com/akutz/goof"
	log "github.com/Sirupsen/logrus"

	"github.com/digitalocean/godo"

	"github.com/codedellemc/libstorage/api/context"
	"github.com/codedellemc/libstorage/api/registry"
	"github.com/codedellemc/libstorage/api/types"

	do "github.com/codedellemc/libstorage/drivers/storage/digitalocean"
	doUtils "github.com/codedellemc/libstorage/drivers/storage/digitalocean/utils"
)
```

Not all my own sources follow this as I've developed this preference over time, but I do try to update files as I work on them. It just makes it easier to scan the imports when looking for specific types of packages.

drivers/storage/digitalocean/storage/digitalocean_storage.go
```diff
@@ -0,0 +1,336 @@
...
+		"token": token,
+	}
+
+	if token == "" {
+		fields["token"] = ""
+	} else {
+		fields["token"] = "******"
+	}
+
+	fields["region"] = d.config.GetString("digitalocean.region")
```
akutz: Hi @protochron,

You should use the constants declared in `github.com/codedellemc/libstorage/drivers/storage/digitalocean` instead of free-form strings in order to prevent possible error. You could even consider writing helper functions such as [those](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/storage/ebs_storage.go#L1321-L1394) found in the EBS driver

drivers/storage/digitalocean/storage/digitalocean_storage.go
```diff
@@ -0,0 +1,336 @@
...
+func (d *driver) Type(ctx types.Context) (types.StorageType, error) {
+	return types.Block, nil
+}
+
+// DigitalOcean volumes are are found using device-by-id, ex:
+// /dev/disk/by-id/scsi-0DO_Volume_volume-nyc1-01 See
+// https://www.digitalocean.com/community/tutorials/how-to-use-block-storage-on-digitalocean#preparing-volumes-for-use-in-linux
+func (d *driver) NextDeviceInfo(
+	ctx types.Context) (*types.NextDeviceInfo, error) {
+	return nil, nil
```
akutz, line 79: Hi @protochron,

If this is not implemented then please return the [appropriate error](https://github.com/codedellemc/libstorage/blob/master/api/types/types_errors.go#L9), ex. `return nil, types.ErrNotImplemented`.

drivers/storage/digitalocean/tests/terraform/.gitignore
```diff
@@ -0,0 +1 @@
+terraform.tfstate*
```
akutz, line 1: Hi @protochron,

I prefer to add entries to the root `.gitignore`, but as long as the entry is relevant to the current directory, that's fine. I've just managed projects in the past where an entry is relevant to some other directory, and the `.gitignore` is removed and the other files get added accidentally.

drivers/storage/digitalocean/utils/client.go
```diff
@@ -0,0 +1,24 @@
...
+
+	client, err := godo.New(oauth2.NewClient(
+		oauth2.NoContext, tokenSrc),
+		godo.SetUserAgent(userAgent()))
+
+	return client, err
+}
+
+func userAgent() string {
+	return "libstorage/" + api.Version.SemVer
```
akutz, line 23: Hi @protochron,

Well damn, I love your idea of the user-agent for libStorage clients. I'm going to steal this. Thanks!

akutz (changes requested):
Hi @protochron,

This looks fantastic. Aside from my inline comments, here are a few additional notes:

* Please try and limit file width to 80 columns. I find it makes for more legible and readable code. That helps promote sustainable sources in my experience.

* You need to add a coverage.mk file to your tests package. Please see the EBS coverage.mk file at https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/tests/coverage.mk for an example.

Thanks again!

akutz:
Hi @protochron,

It would be great to see the addition of a `Vagrantfile` that supports turn-key testing using the DigitalOcean [Vagrant plug-in](https://www.digitalocean.com/community/tutorials/how-to-use-digitalocean-as-your-provider-in-vagrant-on-an-ubuntu-12-10-vps). 

The EBS [`Vagrantfile`](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/tests/Vagrantfile) is a good example of what I mean. It stands up up to three nodes, a libStorage server and two client nodes, and then validates certain operations.

drivers/storage/digitalocean/digitalocean.go
protochron: DO is what we typically use, but that also might be confusing

drivers/storage/digitalocean/storage/digitalocean_storage.go
protochron: Good catch! I thought I fixed all of those

drivers/storage/digitalocean/executor/digitalocean_executor.go
protochron: By that I'm taking it that you mean that `LocalDevices` needs to filter out any mounted drives that are not actually DigitalOcean volumes? That will require https://github.com/digitalocean/godo/pull/122 being merged so that there's an easy API call to check each mounted volume

akutz: Hi @protochron,

Well, the idea is that the executor implementation should not import any non-stdlib dependencies beyond those used broadly, such as some of my support libraries - gofig, gotil, etc. It's to keep the size of the executor to a minimum. LocalDevices doesn't *have* to filter anything -- it's just ideal if it does. For example, for EBS it is able to filter based on the name of the devices, `xvdX`, as opposed to `sdX`, you know?

drivers/storage/digitalocean/digitalocean.go
akutz: Hi @protochron,

I honestly don't mind `do` as long as it doesn't cause a conflict with parsing the YAML (reserved words).

drivers/storage/digitalocean/executor/digitalocean_executor.go
akutz: Hi @protochron,

Do you have time for a quick phone call? I just followed you on Twitter. Follow me and I'll send you my number. I want to explain the purpose of the LocalDevices function. It's easier via voice.

akutz: Hi @protochron,

I sent you my number via a Twitter DM. Please call at your convenience.

drivers/storage/digitalocean/digitalocean.go
protochron: I think we decided on dobs on Slack

akutz (approved):

dcoraboeuf:
Hi,

I'm eager to test the RexRay in DigitalOcean, in order to supply a resilient storage for a Docker Swarm. I'm looking for the documentation on how to configure RexRay for DigitalOcean - do you know where I can find it?

Thanks,
Damien.

kacole2:
@dcoraboeuf you're quick to the draw. the [0.8-RC1](https://github.com/codedellemc/rexray/releases/tag/v0.8.0-rc1) was pushed out yesterday. This part of the libStorage framework which is at [0.5-RC1](https://github.com/codedellemc/libstorage). I just checked the [0.5-RC1 documentation](http://libstorage.readthedocs.io/en/v0.5.0-rc1/user-guide/storage-providers/) and it's missing. I'm also checking the [0.5 milestone](https://github.com/codedellemc/libstorage/milestone/8) and it's not listed. @cduchesne I think this needs to be addressed.

akutz:
Hi @kacole2,

I'm on the hook for the DigitalOcean documentation, so it's my bad. I will address it before GA.

PR #394, opened 2017-01-19
Google Compute Engine (GCE) storage driver

codenrhoden:
This patch re-introduces the GCE persistent disk driver.

Still TODO here:

- [x] docs
- [x] testplan/vagrant/multi-node testing
- [x] tag-based filtering
- [x] supporting SSD-based disks by default

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -589,6 +589,7 @@ remote storage systems. Currently the following storage drivers are supported:
 [EBS](./storage-providers.md#aws-ebs) | ebs, ec2
 [EFS](./storage-providers.md#aws-efs) | efs
 [RBD](./storage-providers.md#ceph-rbd) | rbd
+[GCEPD](./storage-providers.md#gcepd) | gcepd
 ..more coming|
 
 The `libstorage.server.libstorage.storage.driver` property can be used to
--- a/.docs/user-guide/storage-providers.md
+++ b/.docs/user-guide/storage-providers.md
@@ -707,3 +707,107 @@ libstorage:
   in place, it may be possible for a client to attach a volume that is already
   attached to another node. Mounting and writing to such a volume could lead to
   data corruption.
+
+## GCEPD
+
+The Google Compute Engine Persistent Disk (GCEPD) driver registers a driver
+named `gcepd` with the `libStorage` driver manager and is used to connect and
+mount Google Compute Engine (GCE) persistent disks with GCE machine instances.
+
+
+### Requirements
+
+* GCE account
+* Service account credentials in JSON for GCE project. If not using the Compute
+  Engine default service account, create a new service account with the Service
+  Account Actor role, and create/download a new private key in JSON format. see
+  [creating a service account](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#creatinganaccount)
+  for details.
+
+### Configuration
+
+The following is an example with all possible fields configured. For a running
+example see the `Examples` section.
+
+```yaml
+gcepd:
+  keyfile: /etc/gcekey.json
+  zone: us-west1-b
+  defaultDiskType: pd-ssd
+  tag: rexray
+```
+
+#### Configuration Notes
+
+* The `keyfile` parameter is required. It specifies a path on disk to a file
+  containing the JSON-encoded service account credentials. This file can be
+  downloaded from the GCE web portal.
+* The `zone` parameter is optional, and configures the driver to *only* allow
+  access to the given zone. Creating and listing disks from other zones will be
+  denied. If a zone is not specified, the zone from the client Instance ID will
+  be used when creating new disks.
+* The `defaultDiskType` parameter is optional, and specified what type of disk
+  to create, either `pd-standard` or `pd-ssd`. When not specified, the default
+  is `pd-ssd`.
+* The `tag` parameter is optional, and causes the driver to create or return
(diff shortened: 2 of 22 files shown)


codenrhoden:
Okay, I just pushed a commit to add docs. This is ready for "real" review now.

akutz:
HI @codenrhoden,

Please squash and rebase. I believe this has been tested and ready for merge otherwise; is that correct?

cduchesne:
I have tested this driver and it worked quite well for me. Below are a few things I think we should modify before merging

**rexray tag functionality**
I am curious if we're stuck to using the volume's Description field for volume tagging. This field is not editable from the Web UI after a volume has been created. Is this something that you can edit post-creation via api? I'm curious how hard it would be to implement this feature with the GCE Labels feature, which is easily editable from within the Web UI.

**name of driver**
we should name it gcepd (GCE Persistent Disks) instead of gce to avoid potential problems we faced with naming our initial aws ebs driver?

**documentation of api permissions**
we should document which specific api permissions are required to support using this driver

**note limit of persistent disks for smaller instances in documentation**
we should point out that for nodes with 3.75GB or less memory, max # of gce persistent disks is 4 which includes the boot disk. We can link to https://cloud.google.com/compute/docs/disks/ for reference as well.

**10GB size minimum**
we should try to find a way to have REX-Ray spit out a better notification about not making a large enough volume if a user tries to create something under 10GB

codenrhoden:
@cduchesne 

Agree with all your comments. A couple of notes:

re: *rexray tag*. Thanks for showing me that the GCE GUI allows you to use labels, though it is only after disk creation. This implies that they themselves are using the Beta API, as that feature is only in the beta API (and why I didn't use it). Using the standard API, there is no way to modify the description post disk creation. I can use the label feature if we switch to the beta API, which you said you were okay with.

re: *size minimum* this is mostly work we'll have to do in rexray, not libS. Right now libS will return an error of "invalid size" or something similar, but i can make it more specific. I hadn't wanted to hardcode a message about the exact size minimum, but now that I think about it I have to hardcode the size to check against, so it's not a big deal.

codenrhoden:
@cduchesne I've addressed all comments except the rexray tag. Broken out into separate commits so they can be viewed easily. Will work on tags next. once all that is done, will squash down to one commit in preparation for merge.

.docs/user-guide/config.md
```diff
@@ -589,6 +589,7 @@ remote storage systems. Currently the following storage drivers are supported:
 [EBS](./storage-providers.md#aws-ebs) | ebs, ec2
 [EFS](./storage-providers.md#aws-efs) | efs
 [RBD](./storage-providers.md#ceph-rbd) | rbd
+[GCE](./storage-providers.md#gce) | gce
```
akutz: Hi @codenrhoden,

This driver abbreviation still reflects the old `gce` and not the updated `gcepd`.

.docs/user-guide/storage-providers.md
```diff
@@ -707,3 +707,107 @@ libstorage:
   in place, it may be possible for a client to attach a volume that is already
   attached to another node. Mounting and writing to such a volume could lead to
   data corruption.
+
+## GCEPD
+
+The Google Compute Engine (GCE) Persistent Disk driver registers a driver named
```

This driver abbreviation still reflects the old `gce` and not the updated `gcepd`. I'd add a single acronym after the _Persistent Disk_ as (GCEPD). Thoughts?

Makefile
```diff
@@ -1078,8 +1078,18 @@ test-debug:
 	env LIBSTORAGE_DEBUG=true $(MAKE) test
 
 test-rbd:
-	DRIVERS=rbd $(MAKE) deps
-	DRIVERS=rbd $(MAKE) ./drivers/storage/rbd/tests/rbd.test
+	env DRIVERS=rbd $(MAKE) deps
```
akutz: Hi @codenrhoden,

The reason I don't use the `env` command in some places is because of what it does. The `env` command [removes](http://unix.stackexchange.com/a/103474) defined aliases and shell functions on which the build may rely.

akutz (changes requested):

.docs/user-guide/config.md
codenrhoden: Darn, I knew I would miss a place! thanks

.docs/user-guide/storage-providers.md
codenrhoden: I left this one on purpose. But not for a strong reason. I can change it as well.

Makefile
codenrhoden: my inspiration here was the `test-debug` target directly above. What is it about its scenario that makes `env` desirable and the driver-specific test builds not?

akutz: Hi @codenrhoden,

The difference is that I learned more about the `env` command as I worked on this project and only used it at the beginning out of not fully understanding the difference. I rarely use the `test-debug` target and so have never changed it.

.docs/user-guide/storage-providers.md
akutz: Hi @codenrhoden,

It's up to you really. I only pointed it out because it may be the only place we reference GCE? I just want to avoid confusion if possible. If you don't think it's confusing then leave it. It just stood out, but I do not feel strongly about it.

akutz (approved):

cduchesne:
looks good to me, thanks @codenrhoden !

PR #397, opened 2017-01-23
S3FS storage driver

alexey-mr:
This is the draft implementation, or proof-of-concent (POC), of the `s3fs` storage driver that enables the mounting of S3 buckets via FUSE.

## Restrictions
The current implementation has the following restrictions:
* The driver does not support volume creation
* The driver does not support volume removal

## Next Steps
- [ ] Add more options in mounting (e.g. region, now it uses default one)
- [ ] Add ability to mount a sub-path in bucket (bucket:/somepath). In that case volumes are inside of a bucket (allowed in config), e.g. bucket:/volume1, bucket:/volume2, etc. Create|remove volume operations could just create|remove appropriate 'path' in S3 (say, via creation of a hidden object,  e.g. bucket:/volume1/.volume1)
- [ ] Add buckets management via S3 rest api (create|remove in volume create|remove operation). But it is discussable - because buckets are limited resource in S3. Probably it is  not to do it until it is requested by someone.

### DIFF
--- a/drivers/storage/s3fs/executor/s3fs_executor.go
+++ b/drivers/storage/s3fs/executor/s3fs_executor.go
@@ -1,0 +1,224 @@
+// +build !libstorage_storage_executor libstorage_storage_executor_s3fs
+
+package executor
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	gofig "github.com/akutz/gofig/types"
+	"github.com/akutz/goof"
+
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+
+	"github.com/codedellemc/libstorage/drivers/storage/s3fs"
+	"github.com/codedellemc/libstorage/drivers/storage/s3fs/utils"
+)
+
+const (
+	// Template for parsing mount info file (/proc/self/mountinfo)
+	mountinfoFormat = "%d %d %d:%d %s %s %s %s"
+)
+
+// driver is the storage executor for the s3fs storage driver.
+type driver struct {
+	name     string
+	config   gofig.Config
+	credFile string
+}
+
+func init() {
+	registry.RegisterStorageExecutor(s3fs.Name, newDriver)
+}
+
+func newDriver() types.StorageExecutor {
+	return &driver{name: s3fs.Name}
+}
+
+func (d *driver) Init(ctx types.Context, config gofig.Config) error {
+	ctx.Info("s3fs_executor: Init")
+	d.config = config
+	if d.credFile = d.getCredFilePath(); d.credFile == "" {
+		return goof.New(fmt.Sprintf(
+			"%s mount driver requires %s option",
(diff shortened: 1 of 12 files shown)


akutz:
Hi @alexey-mr,

Thank you for this PR. I like the idea of a mount driver; in fact it's been something we've discussed a few times internally due to other driver requirements. Two notes about the way that it's been implemented in this PR:

1. A mount driver should not require its own registry implementation -- it should reuse the central registry.
2. I do not think this should be a driver at all, rather a new function of the executor. 

If it's okay with you, I'd like to go ahead and rewrite the new mount driver for the `s3fs` driver as part of the executor logic. You're free to do so as well, but I also do not mind.

Thoughts so far?

Beyond my comments on the mount driver, I am currently reviewing the rest of the code in the PR. My only other immediate note is that I would like to see the PR's commits squashed and rebased. Thank you.

drivers/os/linux/linux.go
```diff
@@ -67,6 +67,13 @@ func (d *driver) Mounts(
 		return nil, goof.New("cannot specify mountPoint and deviceName")
 	}
 
+	// TODO: s3fs doesnt provide mounted bucket in /proc/self/mountinfo
+	// that's why m.Source is just 's3fs'
+	// so, it is workaround to patch Source info
+	// Another option - use moinPoint to filter mounts into
+	// the integration driver logic and remove this patching
+	d.patchMounts(ctx, &mounts)
```
akutz: Hi @codenrhoden,

Can you please take a look at the suggested `PatchMountInfo` function as well as the implementation of it in this PR and tell me if I'm crazy, but I think we should take the other option @alexey-mr suggests and relocate this to the integration driver [here](https://github.com/cloudscaling/libstorage/blob/95273fdc3263d24c47e244c00db32b34fb55bd5c/drivers/integration/linux/linux.go#L262).

If I'm following all of the code correctly, it appears that the patching/lookup should really be occurring in the `LocalDevices` function of the executor so that the `Libstorage-Localdevices` map is initialized properly when sent to the server and thus the returned volumes have the mappings correctly set. That would mean the correct device name and source should be set when used in the integration driver.

Thoughts?

codenrhoden: @akutz I agree that the patching should happen in the executors `LocalDevices()`. I am not following, however, what would need to be added to the integration driver that is related to this... I feel like I should see it, but I don't.

With your work to add mount/unmount functionality capability to the executors, what would need to be added to the integration driver? I'm assuming a check to see if the volume we are working with needs to be mounted via executor, or one of our existing "standard" mount functions, but is there anything else related to the patching of mount info?

alexey-mr:
Hi @akutz,  I reworked it according updated executor's interface. Now patching of mountinfo.Source is moved in LocalDevices of s3fs executor... But if I understand correctly somethingelse should be done additionally to make it work, because Mounts() (in integration driver) doesnt call LocalDevices from storage driver executors.

akutz:
Hi @alexey-mr,

Thank you again for your work on this. In light of recent changes and in order to close this out, I'm taking ownership of it at #409. Thank you again.

PR #399, opened 2017-01-25
Executor Supported Enhancements

akutz:
This patch enhances the executor CLI to change the behavior of the "supported" function. Internally, storage platforms will still implement the same "Supported" function as previously, but now the executor CLI returns more than simply a flag. The executor CLI will now return a bitmask that corresponds to the functions supported by a storage platform on the executing host. This provides a means to indicate whether an executor supports mounting/unmounting for a given storage platform. The update also centralizes the query for future additions to the executor logic.

This patch also introduces the interface `StorageExecutorWithMount` with two new functions, `Mount` and `Unmount`, but these functions are not yet a part of the storage workflow.

This is a WiP. I will likely squash a future commit onto this PR that allows the new mount functionality in the executor to participate in the standard workflow. I just pushed the PR in order to allow others to review the work so far.

### DIFF
--- a/api/types/types_drivers_executor.go
+++ b/api/types/types_drivers_executor.go
@@ -37,6 +37,12 @@ const (
 	// LSXCmdSupported is the command to execute to find out if an executor
 	// is valid for a given platform on the current host.
 	LSXCmdSupported = "supported"
+
+	// LSXCmdMount is the command for mounting a device to a file system path.
+	LSXCmdMount = "mount"
+
+	// LSXCmdUmount is the command for unmounting mounted file systems.
+	LSXCmdUmount = "umount"
 )
 
 const (
@@ -150,6 +156,23 @@ type StorageExecutorWithSupported interface {
 		opts Store) (bool, error)
 }
 
+// StorageExecutorWithMount is an interface that executor implementations
+// may use to become part of the mount/unmount workflow.
+type StorageExecutorWithMount interface {
+
+	// Mount mounts a device to a specified path.
+	Mount(
+		ctx Context,
+		deviceName, mountPoint string,
+		opts *DeviceMountOpts) error
+
+	// Unmount unmounts the underlying device from the specified path.
+	Unmount(
+		ctx Context,
+		mountPoint string,
+		opts Store) error
+}
+
 // ProvidesStorageExecutorCLI is a type that provides the StorageExecutorCLI.
 type ProvidesStorageExecutorCLI interface {
 	// XCLI returns the StorageExecutorCLI.
@@ -159,7 +182,8 @@ type ProvidesStorageExecutorCLI interface {
 // StorageExecutorCLI provides a way to interact with the CLI tool built with
 // the driver implementations of the StorageExecutor interface.
 type StorageExecutorCLI interface {
-	StorageExecutorWithSupported
+	StorageExecutorFunctions
+	StorageExecutorWithMount
 
 	// WaitForDevice blocks until the provided attach token appears in the
 	// map returned from LocalDevices or until the timeout expires, whichever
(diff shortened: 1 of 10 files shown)


api/types/types_drivers_executor.go
```diff
@@ -150,6 +150,24 @@ type StorageExecutorWithSupported interface {
 		opts Store) (bool, error)
 }
 
+// StorageExecutorWithMount is an interface that executor implementations may
+// may use to become part of the mount/unmount workflow.
```
codenrhoden, line 160: double "may"

cli/lsx/lsx.go
```diff
@@ -75,8 +75,14 @@ func Run() {
...
 				err = opErr
+			} else if opResult {
+				if _, ok := dws.(apitypes.StorageExecutorWithMount); ok {
+					result = apitypes.LSXOpAll
+				} else {
+					result = apitypes.LSXOpAllNoMount
+				}
 			} else {
-				result = opResult
+				result = false
```
codenrhoden: I'd prefer to see this as `LSXSOpNone`, even though they are same value.

drivers/storage/libstorage/libstorage_client_xcli.go
```diff
@@ -19,56 +19,57 @@ import (
...
+		return c.supportedCache.GetLSXSupported(driverName), nil
 	}
 
 	out, err := c.runExecutor(ctx, driverName, types.LSXCmdSupported)
 	if err != nil {
 		if err == types.ErrNotImplemented {
 			ctx.WithField("serviceDriver", driverName).Warn(
 				"supported cmd not implemented")
-			c.supportedCache.Set(driverName, true)
+			c.supportedCache.Set(driverName, types.LSXOpAllNoMount)
```
codenrhoden, line 53: I realize that this code here isn't new, just modified, but I'm surprised it's hear. I thought the containing function here was only called if we had already done a type assertion that the executor implemented the interface that includes the `Supported()` method. As such, I would consider it an actual error if it's defined in the executor, but returns a Not Implemented error. I think this is probably here for backwards compatibility, but is that an issue at this point?

I guess I'm really wondering if it is safe to assume that if `Supported` is not implemented, we say this is `LSXOpAllNomount`, and I believe the answer to that is yes, since that's what we require from a minimal Executor. Which probably means this whole comment is moot. :)

codenrhoden (commented):
Looking good to me! I like the changes.

drivers/storage/libstorage/libstorage_client_xcli.go
akutz: Hi @codenrhoden,

Pretty much :) It's back-compat. The only reason I'm still including it is the outstanding driver PRs there are. I don't know if they've all implemented `Supported` or not. I had originally used this PR to move the `Supported` function to the core type, but as I said, outstanding drivers made me think twice.

api/types/types_drivers_executor.go
```diff
@@ -159,7 +182,8 @@ type ProvidesStorageExecutorCLI interface {
 // StorageExecutorCLI provides a way to interact with the CLI tool built with
 // the driver implementations of the StorageExecutor interface.
 type StorageExecutorCLI interface {
-	StorageExecutorWithSupported
+	StorageExecutorFunctions
+	StorageExecutorWithMount
```
codenrhoden, line 186: Looks like the `StorageExecutorWithSupported` interface is still defined, but not referenced. And it now has the wrong return type of `bool` instead of `LSXSupportedOp` for the `Supported` function. Does it need to stick around, deleted, or updated?

codenrhoden (approved):
Looks good! One question on an unused interface, and that's it.

api/types/types_drivers_executor.go
akutz: As discussed on the phone, there are two interfaces:

1. `StorageExecutorWithSupported`
2. `StorageExecutorCLI`

The former is for the platform implementations of the executor and the latter is for the executor CLI itself. They're both required and correct.

PR #403, opened 2017-02-02
Makefile Fixes / Updates

akutz:
This patch adds the variable `DRIVERS` to the build process where the command `$ DRIVERS=ebs make` would build libStorage with just the EBS storage driver and executor. Additionally, the `RBD_BUILD_TAGS` variable has been removed and the `test-rbd` target has been updated to use the new `DRIVERS` functionality.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,13 +12,14 @@ BUILD_TAGS +=   libstorage_storage_driver \
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


codenrhoden (approved):
👍  👍  👍 

Yes, I'm that enthusiastic about this.

PR #409, opened 2017-02-07
S3FS Fuse Driver Support

akutz:
This patch introduces support for mounting S3FS buckets as Linux file systems. This initial version of the S3FS storage driver does not yet support volume (bucket) creation/removal.

This patch also updates the AWS SDK dependency that is shared by the EBS and EFS storage drivers to v1.6.18, released on 2017/01/27.

Finally, this patch also refactors the `VolumeRemove`, and its related signatures, to use the new `VolumeRemoveOpts` struct instead of a bare `Options` type. All drivers affected by this change have been updated as part of this patch. For people compiling REX-Ray against this branch, you will need the following [patch](https://gist.github.com/akutz/b4c4c839fd7ba745662b7cd2de821a11) in order to do so. The linked patch updates REX-Ray's two sources affected by this change to libStorage.

This PR replaces the existing S3FS PR #397. Thank you @Andrey-mp and @alexey-mr for your work on this, we could not have done this without you.

### DIFF
--- a/drivers/storage/s3fs/executor/s3fs_executor.go
+++ b/drivers/storage/s3fs/executor/s3fs_executor.go
@@ -1,0 +1,209 @@
+// +build !libstorage_storage_executor libstorage_storage_executor_s3fs
+
+package executor
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+
+	log "github.com/Sirupsen/logrus"
+	gofig "github.com/akutz/gofig/types"
+	"github.com/akutz/goof"
+	"github.com/akutz/gotil"
+
+	"github.com/codedellemc/libstorage/api/registry"
+	"github.com/codedellemc/libstorage/api/types"
+
+	"github.com/codedellemc/libstorage/drivers/storage/s3fs"
+	"github.com/codedellemc/libstorage/drivers/storage/s3fs/utils"
+)
+
+// driver is the storage executor for the s3fs storage driver.
+type driver struct {
+	config gofig.Config
+	cmd    string
+	opts   []string
+	szOpts string
+}
+
+func init() {
+	registry.RegisterStorageExecutor(s3fs.Name, newDriver)
+}
+
+func newDriver() types.StorageExecutor {
+	return &driver{}
+}
+
+func (d *driver) Init(ctx types.Context, config gofig.Config) error {
+	d.config = config
+
+	fields := log.Fields{"driver": s3fs.Name}
+	d.cmd = d.config.GetString(s3fs.ConfigS3FSCmd)
+	fields["cmd"] = d.cmd
+
+	if v := d.config.GetStringSlice(s3fs.ConfigS3FSOptions); len(v) > 0 {
+		d.opts = v
+		fields["opts"] = d.opts
+	} else {
+		d.szOpts = d.config.GetString(s3fs.ConfigS3FSOptions)
+		fields["opts"] = d.szOpts
+	}
+
+	ctx.WithFields(fields).Debug("storage executor initialized")
+	return nil
+}
+
+func (d *driver) Name() string {
+	return s3fs.Name
+}
+
+// Supported returns a flag indicating whether or not the platform
+// implementing the executor is valid for the host on which the executor
+// resides.
+func (d *driver) Supported(
+	ctx types.Context,
+	opts types.Store) (bool, error) {
+
+	return gotil.FileExistsInPath(d.cmd), nil
+}
+
+// InstanceID
+func (d *driver) InstanceID(
+	ctx types.Context,
+	opts types.Store) (*types.InstanceID, error) {
+	return utils.InstanceID(ctx, d.config)
+}
+
+// NextDevice returns the next available device.
+func (d *driver) NextDevice(
+	ctx types.Context,
+	opts types.Store) (string, error) {
+	return "", types.ErrNotImplemented
+}
+
+// Return list of local devices
+func (d *driver) LocalDevices(
+	ctx types.Context,
+	opts *types.LocalDevicesOpts) (*types.LocalDevices, error) {
+
+	buckets, err := d.getMountedBuckets(ctx)
+	if err != nil {
+		return nil, err
+	}
+	return &types.LocalDevices{Driver: s3fs.Name, DeviceMap: buckets}, nil
+}
+
+func (d *driver) Mount(
+	ctx types.Context,
+	deviceName, mountPoint string,
+	opts *types.DeviceMountOpts) error {
+
+	if mp, ok := d.findMountPoint(ctx, deviceName); ok {
+		ctx.WithFields(log.Fields{
+			"bucket":     deviceName,
+			"mountPoint": mp,
+		}).Debug("bucket is already mounted")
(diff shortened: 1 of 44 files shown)


drivers/storage/s3fs/executor/s3fs_executor.go
```diff
@@ -0,0 +1,187 @@
...
+	ctx types.Context,
+	bucket, mountPoint string,
+	opts *types.DeviceMountOpts) error {
+
+	args := []string{bucket, mountPoint}
+	if len(d.opts) > 0 {
+		for _, o := range d.opts {
+			args = append(args, fmt.Sprintf("-o%s", o))
+		}
+	} else if d.szOpts != "" {
```
codenrhoden, line 153: I'm confused by the difference between `opts` and `szOpts`. I mean, I can see that the former is a string, and the latter is a slice of strings, but why are the two necessary? The docs say that `options` is supposed to be a list of options, and that `-o` should not be included in the config value. So shouldn't only one of these be necessary?

Perhaps I'm not familiar with `gofig`, but I don't know why, in the `Init` function, both `GetString` and `GetStringSlice` must be used. But if they _do_ have to be used, would it be possible to just have one "opts" field in the driver stuct, as a string slice, and put in the single entry rather than having two different opts fields?

What does "sz" in `szOpts` mean here, Size? that's how I read it. But I also don't get why the `-o` isn't required.

drivers/storage/s3fs/s3fs.go
```diff
@@ -0,0 +1,128 @@
...
+	r.Key(gofig.String,
+		"",
+		"",
+		"Tag prefix for S3 naming",
+		ConfigS3FSTag)
+	r.Key(gofig.Int,
+		"",
+		DefaultMaxRetries,
+		"Max number of times to retry failed operations",
+		ConfigS3FSMaxRetries)
```
codenrhoden, line 123: The hostname, tag, and maxretries options are not mentioned in the docs.

drivers/storage/s3fs/storage/s3fs_storage.go
```diff
@@ -0,0 +1,392 @@
...
+
+// NextDeviceInfo returns the information about the driver's next available
+func (d *driver) NextDeviceInfo(
+	ctx types.Context) (*types.NextDeviceInfo, error) {
+	return s3fsUtils.NextDeviceInfo, nil
+}
+
+// Type returns the type of storage the driver provides.
+func (d *driver) Type(ctx types.Context) (types.StorageType, error) {
+	return types.Object, nil
```
codenrhoden, line 200: Is this true? Even though the backing storage is object, you end up with a file system, so I would think the type is `File`.

drivers/storage/s3fs/storage/s3fs_storage.go
```diff
@@ -0,0 +1,392 @@
...
+		return goof.WithFieldE(
+			"volumeID", volumeID, "error deleting s3 bucket", err)
+	}
+
+	return nil
+}
+
+// VolumeAttach attaches a volume and provides a token clients can use
+// to validate that device has appeared locally.
+func (d *driver) VolumeAttach(
```
codenrhoden, line 470: Can we add  a comment that for s3fs, the attach is not performed server-side, but is left to the executor?

drivers/storage/s3fs/storage/s3fs_storage.go
```diff
@@ -0,0 +1,392 @@
...
+		ID:   bucket,
+	}
+
+	if attachments.Requested() {
+		vatt := &types.VolumeAttachment{
+			VolumeID:   bucket,
+			DeviceName: bucket,
+		}
+		if iid, iidOK := context.InstanceID(ctx); iidOK {
+			vatt.InstanceID = iid
```
codenrhoden: I know things were weird with this driver, but isn't this going to list every single bucket as attached to the instance if a client provides the IID header?

drivers/storage/s3fs/utils/utils.go
```diff
@@ -0,0 +1,25 @@
...
+
+// InstanceID returns the instance ID for the local host.
+func InstanceID(
+	ctx types.Context, config gofig.Config) (*types.InstanceID, error) {
+
+	var hostName string
+	if config == nil {
+		hostName = config.GetString(s3fs.ConfigS3FSHostName)
+	} else {
+		hostName, _ = os.Hostname()
```
codenrhoden, line 22: It wasn't until I looked at this code that it dawned on me that this driver is useful outside of AWS. Of course it is, but I just hadn't thought about it. Might be something worth pointing out in the docs, because it is really useful?

Would there be any merit it trying to make sure there is a more uniqueness to the IID, by appending a MAC address or anything? Probably not, but I think once you jump out of the cloud provider, you are much more likely to run into hostname collisions.

akutz:
HI @codenrhoden,

I'm going to call you to discuss many of your comments due to the unique nature of this driver.

drivers/storage/s3fs/executor/s3fs_executor.go
akutz: Hi @codenrhoden,

So, `szOpts` stands for _string of characters terminated by null character_ ... Opts. The `sz` prefix is often found in use by C or C++ programmers to denote a null-terminated string. I will sometimes use it to indicate a variable is a string if there is a similarly named variable of a different type.

The properties in the config file don't really have types per se. It's more how they're interpreted upon being consumed. Thus while it's cleaner to specify the options as an array of `arg=value` or `flag` items, maybe I also want to read it as a string? The above code first attempts to parse the `s3fs.options` value as an array of strings. If it isn't it reads it as a string. That way a user can determine how to put in the options.

drivers/storage/s3fs/s3fs.go
akutz: Hi @codenrhoden,

They're really developer options for overriding the instance ID value and for debugging.

drivers/storage/s3fs/storage/s3fs_storage.go
akutz: Hi @codenrhoden,

This value refers to the backing storage type.

drivers/storage/s3fs/storage/s3fs_storage.go
akutz: Hi @codenrhoden,

The client performs a mount, not an attach.

drivers/storage/s3fs/storage/s3fs_storage.go
akutz: Hi @codenrhoden,

Ah, you're right. I should make it so all volumes are available if the instance ID is not provided. Thanks.

drivers/storage/s3fs/utils/utils.go
akutz: HI @codenrhoden,

Yes, the S3FS driver is not at all required to be used on EC2 instances. I'll add that to the documentation if you think it is important to call it out. I figured anyone wanting to use S3 would understand that, but there's no harm in calling it out.

The hostname and instance ID are not used by the storage driver except for logging. Even there it could be useful for them to be unique I suppose.

PR #410, opened 2017-02-10
ScaleIO and EBS Automated Testing

davidvonthenen:
Verified on 2 environments (Chris' config and AWS). You can run on bare metal or stand up a test environment in AWS using a provided CloudFormation template.

**To run in AWS**
The test-env-up.sh sets up the environment in AWS. AWS was chosen because the entire scaleio environment can be propped up quickly (in under a minute). An AWS key/secret will be required with sufficient permissions. A uniquely identifiable stack-name must be given which allows for the process to stand up multiple environments concurrently.

The test-run.sh pushes over required files, the test script containing multiple tests, and then executes the test script. Once the test cases are completed, an output file containing the test results are copied back to the system.

The test-env-down.sh script is run last and removes all artifacts in Amazon.

**To run on bare metal**
You can use the test-run-metal.sh. It requires 3 nodes running the ScaleIO sdc, go, and jq packages. Will execute the test cases above on those nodes. This is just a "nice to have" and not the focus of the PR.

### DIFF
--- a/drivers/storage/ebs/tests/config-tmpl.yml
+++ b/drivers/storage/ebs/tests/config-tmpl.yml
@@ -1,0 +1,18 @@
+rexray:
+  logLevel: warn
+  modules:
+    rexray:
+      type: docker
+      libstorage:
+        service: ebs
+      host: unix:///run/docker/plugins/rexray.sock
+libstorage:
+  service: ebs
+  integration:
+    volume:
+      operations:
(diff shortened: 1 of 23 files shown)


davidvonthenen:
This addresses issues https://github.com/codedellemc/rexray/issues/712 and https://github.com/codedellemc/rexray/issues/713

Issue #411, opened 2017-02-10
No Ceph Monitors found


tuxity:
Hi!

I followed https://github.com/dcos/examples/tree/master/1.8/ceph to create a ceph cluster and everything works great.

But using rexray with it, I had the following error:
```
error: error getting instance ID: No Ceph Monitors found
```

It seems to be because of my [ceph.conf](https://github.com/dcos/examples/tree/master/1.8/ceph#cephconf). The command generates me this line:
```
mon host = "mon-dxrzn-s16.ceph.mesos.:1025,mon-etctw-s15.ceph.mesos.:1025"
```

tuxity:
It's working thanks to #412 👍

PR #412, opened 2017-02-10
Fix for Ceph MonIPs with Ports

akutz:
This patch is a first-attempt at fixing a bug where the RBD executor does not properly parse CephMon addresses with ports.

This patch does not account for IPv6 addresses as those require brackets for parsing, but this should hopefully help a community user move forward sooner than later.

### DIFF
--- a/drivers/storage/rbd/executor/rbd_executor.go
+++ b/drivers/storage/rbd/executor/rbd_executor.go
@@ -160,17 +160,19 @@ func getCephMonIPs() ([]net.IP, error) {
 
 	monStrings := strings.Split(strings.TrimSpace(string(out)), ",")
 
-	monIps := make([]net.IP, 0, 4)
+	monIps := []net.IP{}
 
 	for _, mon := range monStrings {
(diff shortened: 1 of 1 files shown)


codenrhoden:
I personally don't see the need to do more than this. I know you are worried about brackets with IPv6, but if it's an improperly formatted IPv6 address, I don't think we should try to fix it.

I double-checked that `LookupIP` will just return the IP if it's passed an IP and it does (which I'm sure you knew).

I'd say run with this, and no need to do more at this time.

akutz:
Hi @codenrhoden,

IPv6 addresses do not *require* brackets except for some situations. If the CephMon IP list requires IPv6 addresses to use brackets, then we're square. If it doesn't, or worse, doesn't support brackets, then we'd need to handle that, right?

akutz:
Hi @codenrhoden,

I'm merging this for now, and if we want to augment later, then great.

akutz:
Closes #411.

PR #413, opened 2017-02-11
Fix for Premption and Volume Mounting

akutz:
This patch updates the Linux integration driver's preemption logic when mounting volumes. If a volume is already attached the branch in the Mount function involving a precautionary unmount is not invoked.


akutz:
Hi @cduchesne & @clintonskitson,

akutz:
Hi @cduchesne & @clintonskitson,

Please feel free to update the PR's description or add a comment to indicate the background on this.

PR #414, opened 2017-02-14
ScaleIO Executor Enhancements

akutz:
This patch enhances the ScaleIO executor so that the `drv_cfg` binary is no longer hard-coded and can now be configured with the configuration property `scaleio.drvCfg` which translates into the environment variable `SCALEIO_DRVCFG`. 

If the property `scaleio.guid` (env var `SCALEIO_GUID`) is set then the `drv_cfg` binary is not used at all to discover the client's GUID.

This patch also updates the logic to discover local devices so the `drv_cfg` is no longer necessary or required to provide this information.

### DIFF
--- a/drivers/storage/scaleio/executor/scaleio_executor.go
+++ b/drivers/storage/scaleio/executor/scaleio_executor.go
@@ -3,28 +3,27 @@
 package executor
 
 import (
-	"fmt"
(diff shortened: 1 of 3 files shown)


codenrhoden:
👍  This looks a lot cleaner, too!

PR #416, opened 2017-02-15
Implement GCE application default credentials

codenrhoden:
Enhance the GCEPD driver by adding support for application default
credentials. With this patch, A user no longer has to upload or provide
a JSON encoded file with service account credentials, as the GCE client
library will automatically fetch any service account credentials
associated with the GCE instances via the metadata server.

Improve docs to clarify what permissions are required of a service
account, regardless of whether you are providing it via JSON or
metadata lookup.

### DIFF
--- a/.docs/user-guide/storage-providers.md
+++ b/.docs/user-guide/storage-providers.md
@@ -708,7 +708,7 @@ libstorage:
   attached to another node. Mounting and writing to such a volume could lead to
   data corruption.
 
-## GCEPD
+## GCE Persistent Disk
 
 The Google Compute Engine Persistent Disk (GCEPD) driver registers a driver
 named `gcepd` with the `libStorage` driver manager and is used to connect and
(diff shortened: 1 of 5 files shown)


.docs/user-guide/storage-providers.md
```diff
@@ -739,36 +743,45 @@ gcepd:
...
+  `keyfile` is *not* specified, the application will try to lookup
+  [application default credentials](https://developers.google.com/identity/protocols/application-default-credentials).
+  This has the effect of looking for credentials in the priority described
+  [here](https://godoc.org/golang.org/x/oauth2/google#FindDefaultCredentials).
 * The `zone` parameter is optional, and configures the driver to *only* allow
   access to the given zone. Creating and listing disks from other zones will be
   denied. If a zone is not specified, the zone from the client Instance ID will
   be used when creating new disks.
-* The `defaultDiskType` parameter is optional, and specified what type of disk
+* The `defaultDiskType` parameter is optional, and specifies what type of disk
```

This should omit the comma as follows:

> The `defaultDiskType` parameter is optional and specifies what type of disk

The secondary part of the sentence is not a sentence on its own and thus a conjunctive comma is inappropriate.

.docs/user-guide/storage-providers.md
```diff
@@ -718,10 +718,14 @@ mount Google Compute Engine (GCE) persistent disks with GCE machine instances.
 ### Requirements
 
 * GCE account
-* Service account credentials in JSON for GCE project. If not using the Compute
-  Engine default service account, create a new service account with the Service
-  Account Actor role, and create/download a new private key in JSON format. see
-  [creating a service account](https://developers.google.com/identity/protocols/OAuth2ServiceAccount#creatinganaccount)
+* The libStorage server must be running on a GCE instance created with a Service
```

@cantbewong suggested the following change:

> The libStorage server must be running on a GCE instance created with a Service Account with appropriate **permissions**, or a Service Account credentials file in JSON format must be supplied. If not using the Compute Engine default Service Account with the Cloud Platform/"all cloud APIs" scope, create a new Service Account via the IAM portal **(Service accounts tab)**. This Service Account requires the **Compute Engine/Instance Admin, Compute Engine/Storage Admin, and Project/Service Account Actor** roles. Then create/download a new private key in JSON format. see creating a service account for details. **Also note that if permissions on a service account are edited, an instance must be restarted to have them take effect.**

I'm noting a change to Steve's suggestion:

>  The REX-Ray service must be restarted in order for permissions changes on a service account to take effect.

cantbewong: yes, agree with @akutz

codenrhoden:
Suggestions implemented. I linked directly to the Service Accounts page within the IAM portal instead of telling you to navigate there.

I also mentioned restarting libStorage instead of REX-Ray -- I try not to reference REX-Ray from libStorage docs.

cantbewong:
LGTM

PR #420, opened 2017-02-16
Service Names & Instance IDs

akutz:
This patch updates the libStorage framework logic to store instance IDs in maps with the service name as the key instead of the driver name. This change is to account for the issue codedellemc/rexray#685.

The instance ID format when marshaled as a string has been updated to the following:

```
DRIVER[:SERVICE]=ID,FIELDS,METADATA
```

This patch is backwards compatible with regards to the published REST API. If an instance ID header does not include an optional `SERVICE` component, the instance ID will be used for all services with the specified driver type unless another instance ID header with a specified service overrides it.


codenrhoden (approved):
👍

PR #421, opened 2017-02-17
Azure Unmanaged Disk support

codenrhoden:
This patchset builds off of work from @Andrey-mp in #372, and supersedes that PR.

Two main things missing right now are:
 * re-verifying tests (they were written, but I haven't run/updated them yet)
 * documentation

The documentation for this will be important. Setting up the required accounts and permissions is not trivial, and I found it complicated for myself.

Another issue that we may run into is that Azure is very slow when it comes to disk operations. Just throwing raw API requests (via `curl`) at libStorage, I often saw attach/detach requests take 50-80 seconds. The task timeout in libS seems to be 60 seconds, at which point a task ID is returned that you can check the status of.  Not sure how this will play with REX-Ray.


codenrhoden:
Tests pass now!  I did have to run it multiple times to avoid the 60sec timeout issue for volume attach/detach tests...  finally got it.

```
[trhoden@test1 ~]$ ./azuredisk.test | tee testoutput.txt
PASS
coverage: 21.2% of statements in github.com/codedellemc/libstorage/drivers/storage/azuredisk, github.com/codedellemc/libstorage/drivers/storage/azuredisk/executor
```

codenrhoden:
Okay, missing docs, but I think this is good to go for a review and see if it can be included in -RC2.

akutz (approved):

gpaul:
> I often saw attach/detach requests take 50-80 seconds. The task timeout in libS seems to be 60 seconds, at which point a task ID is returned that you can check the status of. Not sure how this will play with REX-Ray.

Any update on this? Specifically: I saw REX-Ray updated to v0.5.2 of libstorage recently but I couldn't find any such async retry logic.

codenrhoden:
@gpaul The [documentation](http://libstorage.readthedocs.io/en/stable/user-guide/storage-providers/#runtime-behavior_5) for the Azure driver tells you how to work around this.

In short, bump that default 60 second timeout to something greater, like 120.

gpaul:
@codenrhoden Thanks for the link to the docs. Does Azure provide a <120s SLA? It seems like this will just makes failures less common and therefore more devestating. It seems like libstorage does exactly the right thing by allowing async operation, the only question is whether eg., REX-Ray takes proper advantage of that. I'll open a RR issue, thanks.

PR #422, opened 2017-02-17
Executor CLI Volume Enhancements

akutz:
This path adds the following commands to the executor:

* volumeCreate
* volumeRemove
* volumeAttach
* volumeDetach

Now storage platforms can participate in the above workflow logic on the client-side without being part of the libStorage client.

### DIFF
--- a/api/types/types_drivers_executor.go
+++ b/api/types/types_drivers_executor.go
@@ -46,6 +46,18 @@ const (
 
 	// LSXCmdMounts is the command for getting a list of mount info objects.
 	LSXCmdMounts = "mounts"
+
+	// LSXCmdVolumeCreate is the command for creating a volume.
+	LSXCmdVolumeCreate = "volumeCreate"
+
+	// LSXCmdVolumeRemove is the command for removing a volume
+	LSXCmdVolumeRemove = "volumeRemove"
+
+	// LSXCmdVolumeAttach is the command for attaching a volume.
+	LSXCmdVolumeAttach = "volumeAttach"
+
+	// LSXCmdVolumeDetach is the command for detaching a volume.
+	LSXCmdVolumeDetach = "volumeDetach"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


api/types/types_drivers_executor.go
```diff
@@ -46,6 +46,18 @@ const (
...
+	LSXCmdVolumeCreate = "volumeCreate"
+
+	// LSXCmdVolumeRemove is the command for removing a volume
+	LSXCmdVolumeRemove = "volumeRemove"
+
+	// LSXCmdVolumeAttach is the command for attaching a volume.
+	LSXCmdVolumeAttach = "volumeAttach"
+
+	// LSXCmdVolumeDeatch is the command for detaching a volume.
+	LSXCmdVolumeDeatch = "volumeDetach"
```
codenrhoden: s/VolumeDeatch/VolumeDetach/

codenrhoden (changes requested):
Looks good, but seeing this type makes me realize that none of the new consts created, `LSXCmdVolume{Create,Remove,Attach,Detach}` are actually used anywhere. What am I missing?

api/types/types_drivers_executor.go
akutz: Hi @codenrhoden,

Sorry, this is labelled a work-in-progress, but I only updated the situation in Slack where I mentioned the PR is incomplete.

codenrhoden: Ah yes, it clearly does say that. Apologies. That would be the answer to "what am I missing"?  😄

akutz: image

akutz: Hi @codenrhoden,

Sorry, I wasn't trying to be snarky. I had not refreshed the page and did not see your follow-up. The kids do not have school, and I'm busy watching She-Ra with my daughter while my son is at speech therapy. :)

akutz:
Hey @codenrhoden,

Now the executor should be complete.

akutz:
As this PR isn't required for this release, I will be changing this PR's base to `master` once libStorage 0.5.0 is GA.

PR #424, opened 2017-02-21
Ceph monitor port support (IPv4/IPv6)

codenrhoden:
Came across this during early testing for 0.5.0-RC2. Without this fix, any use of the rbd driver on clusters *without* a monitor port defined fails, which is likely the majority. This fixes a regression introduced by 1040f4c.

Previous fix to support Ceph monitors with non standard ports caused a
new issue for monitors *without* ports. The call to net.SplitHostPort()
fails if the string does not contain a ":".

Rework the parsing of monitor addresses to account for the
presence/absence of ports, both for IPv4 and IPv6. As it turns out, The
Ceph config file also requires that IPv6 addresses (but not ports!) be
enclosed by brackets, which is the same requirement for
net.SplitHostPort(), and is something taken advantage of in this patch.

Tests added to cover possible combinations. Tests pass (and in fact, made me rework this solution several times!)

### DIFF
--- a/drivers/storage/rbd/executor/rbd_executor.go
+++ b/drivers/storage/rbd/executor/rbd_executor.go
@@ -160,20 +160,9 @@ func getCephMonIPs() ([]net.IP, error) {
 
 	monStrings := strings.Split(strings.TrimSpace(string(out)), ",")
 
-	monIps := []net.IP{}
-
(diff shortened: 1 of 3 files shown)


akutz:
HAHAHAHAHAHAHAHAHAHAHAHAHAHAHAHA!

PR #425, opened 2017-02-21
Allow dashes in RBD names and pools

codenrhoden:
Document allowed characters, and return a better error message when
invalid characters are found


akutz:
Please rebase this onto `release/0.5.0` and/or remove the Ceph MonIP commit from this PR as it's been merged.

PR #430, opened 2017-02-22
Updated Developer Documentation

akutz:
This patch updates the project's developer documentation.

### DIFF
--- a/.docs/dev-guide/build-reference.md
+++ b/.docs/dev-guide/build-reference.md
@@ -4,7 +4,63 @@ How to build libStorage
 
 ---
 
-## Build Requirements
+## Basic Builds
+The following one-line command is the quickest, simplest, and most
+deterministic approach to building libStorage:
+
+```bash
+$ git clone https://github.com/codedellemc/libstorage && make -C libstorage
+```
+
+!!! note "note"
+
+    The above `make` command defaults to the `docker-build` target only if
+    Docker is detected and running on a host, otherwise the `build` target is
+    used. For more information about the `build` target, please see the
+    [Advanced Builds](#advanced-builds) section.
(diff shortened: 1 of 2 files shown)


.docs/dev-guide/build-reference.md
```diff
@@ -4,7 +4,56 @@ How to build libStorage
...
 ---
 
-## Build Requirements
+## Basic Builds
+The following one-line command is the quickest, simplest, and most
+deterministic approach to building libStorage:
+
+```bash
+$ git clone https://github.com/codedellemc/libstorage && make -C libstorage
+```
```
codenrhoden, line 13: I think the only thing I see missing is an explanation that if Docker is present, it will be used by default (even if you just do `make` and not `make docker-build`. If Docker is not present, it builds locally, just like `make build`.

codenrhoden (changes requested):
Nice addition, @akutz.  Really clear and well thought out. Just the one comment.

.docs/dev-guide/build-reference.md
akutz: Hi @codenrhoden,

I thought about that. I even had a brief explanation under the command with what it does exactly, but I didn't like the way it crunched the formatted code between two paragraphs. It was distracting visually. I may add another section underneath that with a proviso or use the note formatting block to indicate the explanation.

akutz: Hi @codenrhoden,

Thoughts on this type of note?

image

codenrhoden: 👍

codenrhoden (approved):
Thanks for adding that note!

PR #470, opened 2017-03-20
Scripts for testing EBS, EFS, and ScaleIO

davidvonthenen:
Scripts for testing EBS, EFS, and ScaleIO. These scripts are for user-triggered testing. The scripts will stand up the environments in AWS, run the test cases, and then clean up the environment. This satisfies the following Issue https://github.com/codedellemc/libstorage/issues/131.

The test-env-up.sh sets up the environment in AWS. AWS was chosen because the entire scaleio environment can be propped up quickly (in under a minute). An AWS key/secret will be required with sufficient permissions. A uniquely identifiable stack-name must be given which allows for the process to stand up multiple environments concurrently.

The test-run.sh pushes over or builds required files, the test script containing multiple tests, and then executes the test script. Once the test cases are completed, an output file containing the test results are copied back to the system.

The test-env-down.sh script is run last and removes all artifacts in Amazon.


akutz:
Hi @dvonthenen,

I've received this and have started to review it. It may take a bit. Thank you!

davidvonthenen:
Thanks. Just to make it clear, this is for offline user-triggered testing only. The integrated CI testing is still a work in progress but will reuse a lot of these components.

PR #479, opened 2017-03-21
Parse Request Opts Config Option

akutz:
This patch defines a new configuration property `libstorage.server.parseRequestOpts`. When this property is set to true, incoming POST requests will be parsed for an opts field, and key/value pairs in the opts field's map will be promoted to the request proper.

For example, the following volume creation POST request bodies are now equivalent when the above property is enabled:

## Classic Request
```json
{
    "name":      "newVolume",
    "encrypted": true
}
```

## With Opts Request
```json
{
    "name": "newVolume",
    "opts": {
        "encrypted": true
    }
}
```

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -921,3 +921,60 @@ libstorage:
         mount:
           rootPath: /data
 ```
+
+### REST Configuration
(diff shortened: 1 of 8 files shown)


codenrhoden (approved):
LGTM

PR #481, opened 2017-03-21
Fix for Linux int driver omitting encryption

akutz:
This patch updates the Linux integration driver so that it processes volume creation requests with encryption properly.

### DIFF
--- a/drivers/integration/linux/linux.go
+++ b/drivers/integration/linux/linux.go
@@ -5,7 +5,6 @@ import (
 	"strings"
 
 	"fmt"
-	"strconv"
 
 	log "github.com/Sirupsen/logrus"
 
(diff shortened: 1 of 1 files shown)


drivers/integration/linux/linux.go
```diff
@@ -459,8 +459,10 @@ func (d *driver) Create(
 	volumeType := d.volumeType()
 	optsNew.Type = &volumeType
 	io, _ := strconv.Atoi(d.iops())
```
codenrhoden: I know it's not part of this PR, but I hate seeing this err suppressed. It doesn't appear that the syntax for this config value has been validated previously, so it's possible this conversion should fail. Something worth addressing?

akutz: I concur. In a separate PR, but definitely.

akutz: Hi @codenrhoden,

I addressed this concern in a rebase. Care to review?

PR #485, opened 2017-03-24
Always return err if volume is not found

### LINKED ISSUE #396: VolumeInspect crashes for some drivers if no volume is found (nil deref)
Hey @akutz, I need your expertise on what the right way to fix this is...

I noticed that if `VolumeInspect` returns `nil, nil`, the framework crashes with a nil deref. Is it always expected that the returned volume will be non-nil if `err != nil`? Put another way, if the volume is not found, should we return an error?

The drivers are not consistent about this. When I'm writing new drivers, I've been looking mostly at EBS and ScaleIO as guides, but they take different approaches.

EBS: (https://github.com/codedellemc/libstorage/blob/master/drivers/storage/ebs/storage/ebs_storage.go#L273-L275)
```
	if len(ec2vols) == 0 {
		return nil, errNoVolReturned
	}
```

ScaleIO (https://github.com/codedellemc/libstorage/blob/master/drivers/storage/scaleio/storage/scaleio_storage.go#L273-L275)
```
	if len(volumes) == 0 {
		return nil, nil
	}

```

If I'm reading things correctly, the ScaleIO driver would crash if you did an InstanceInspect on a non-existing volume. Or perhaps it doesn't because the underlying API returns an error, which gets caught first, and in that case the code I'm referencing is dead code. Either way, it's confusing.  Here is a partial backtrace of the crash:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x6bbdb6]

goroutine 29 [running]:
panic(0xab3300, 0xc4200140b0)
        /usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/codedellemc/libstorage/api/server/router/volume.handleVolAttachments(0xed9200, 0xc42004b800, 0x0, 0xc420294680, 0x0, 0x0, 0x0)
        /go/src/github.com/codedellemc/libstorage/api/server/router/volume/volume_routes.go:139 +0x446
github.com/codedellemc/libstorage/api/server/router/volume.(*router).volumeInspect.func2(0xed9200, 0xc42004b800, 0xed3740, 0xc42019ce00, 0xc42035d270, 0xc420297e00, 0xb95802, 0xc)
        /go/src/github.com/codedellemc/libstorage/api/server/router/volume/volume_routes.go:353 +0x178
```

The line numbers match up fine with the current `master`, and it is simply because `InstanceInspect` returns `nil,nil`.

Super easy fix whichever direction we go, just need your guidance.

Depending on your answer, I'm happy to fix this either by correcting the drivers (where necessary), or adding a check in volume_routes to check for nil.

codenrhoden:
From VolumeInspect, it is not allowed to return nil, nil if the
referenced volume is not found. However, several drivers do so, and this
can lead to a nil dereference.

Fixes #396

### DIFF
--- a/drivers/storage/isilon/storage/isilon_storage.go
+++ b/drivers/storage/isilon/storage/isilon_storage.go
@@ -16,6 +16,7 @@ import (
 	"github.com/codedellemc/libstorage/api/context"
 	"github.com/codedellemc/libstorage/api/registry"
 	"github.com/codedellemc/libstorage/api/types"
+	apiUtils "github.com/codedellemc/libstorage/api/utils"
 	"github.com/codedellemc/libstorage/drivers/storage/isilon"
 )
 
@@ -245,7 +246,7 @@ func (d *driver) VolumeInspect(
 	}
 
 	if vols == nil {
-		return nil, nil
+		return nil, apiUtils.NewNotFoundError(volumeID)
 	}
 
 	return vols[0], nil
@@ -349,9 +350,6 @@ func (d *driver) VolumeAttach(
 	if err != nil {
 		return nil, "", err
 	}
-	if vol == nil {
-		return nil, "", goof.New("no volumes returned")
-	}
 
 	exportID, err := d.client.ExportVolume(ctx, volumeID)
 	if err != nil {
@@ -423,14 +421,10 @@ func (d *driver) VolumeDetach(
 	d.Lock()
 	defer d.Unlock()
 
-	vol, err := d.VolumeInspect(ctx, volumeID,
+	_, err := d.VolumeInspect(ctx, volumeID,
 		&types.VolumeInspectOpts{Attachments: 0})
 	if err != nil {
 		return nil, err
-	}
-
-	if vol == nil {
-		return nil, goof.New("no volumes returned")
 	}
 
 	instanceID, err := d.InstanceInspect(ctx, nil)
--- a/drivers/storage/vbox/storage/vbox_storage.go
+++ b/drivers/storage/vbox/storage/vbox_storage.go
@@ -17,6 +17,7 @@ import (
 	"github.com/codedellemc/libstorage/api/context"
 	"github.com/codedellemc/libstorage/api/registry"
 	"github.com/codedellemc/libstorage/api/types"
+	apiUtils "github.com/codedellemc/libstorage/api/utils"
 	"github.com/codedellemc/libstorage/drivers/storage/vbox"
 )
 
@@ -163,7 +164,7 @@ func (d *driver) getVolumeMapping(ctx types.Context) ([]*types.Volume, error) {
 		return nil, err
 	}
 
-	if err := m.Refresh(); err != nil {
+	if err = m.Refresh(); err != nil {
 		return nil, err
 	}
 	defer m.Release()
@@ -336,14 +337,14 @@ func (d *driver) VolumeAttach(
 	}
 
 	if len(volumes) == 0 {
-		return nil, "", goof.New("no volume found")
+		return nil, "", apiUtils.NewNotFoundError(volumeID)
 	}
 
 	if len(volumes[0].Attachments) > 0 && !opts.Force {
 		return nil, "", goof.New("volume already attached to a host")
 	}
 	if opts.Force {
-		if _, err := d.VolumeDetach(ctx, volumeID, nil); err != nil {
+		if _, err = d.VolumeDetach(ctx, volumeID, nil); err != nil {
 			return nil, "", err
 		}
 	}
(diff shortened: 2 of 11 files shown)


codenrhoden:
@akutz. What do you think of the use of `goof.WithField()` here? A little different than what the other drivers have done...

In fact, I think we could probably standardize across all the drivers. I see one outlier still, and I'm not sure if it's the right use or not.  See [EFS](https://github.com/codedellemc/libstorage/blob/v0.5.1/drivers/storage/efs/storage/efs_storage.go#L351-L353), it does something that no other driver does, and uses `types.ErrNotFound{}`.  From what I can see of that [type](https://github.com/codedellemc/libstorage/blob/v0.5.1/api/types/types_errors.go#L23), returning that as an empty struct isn't all that useful.  The only place where I can find that type being used is [here](https://github.com/codedellemc/libstorage/blob/v0.5.1/api/utils/utils_errors.go#L28-L32). At a minimum, that's probably the syntax EFS should use (initializing the `Goof` field).

Would returning this concrete type be helpful here? If so, I can change all drivers to do that instead.  Whatever the answer is, I'd like them all to be the same. Even with this change, there are at least three ways of returning information about this "volume not found" error.

codenrhoden:
@akutz Okay, I updated a lot more things to use `NewNotFoundError()`. Going through these drivers, I find a lot of refactoring I want to do. :)

There's a couple places where I removed wrapping an existing error in a `goof.WithError()`, because that would trip up the error handling middleware.

drivers/storage/isilon/storage/isilon_storage.go
```diff
@@ -349,9 +350,6 @@ func (d *driver) VolumeAttach(
 	if err != nil {
 		return nil, "", err
 	}
-	if vol == nil {
-		return nil, "", goof.New("no volumes returned")
-	}
```
codenrhoden, line 354: This code isn't necessary because `d.VolumeInspect` will already return an error if vol is nil. So this is dead code.

codenrhoden:
@akutz I believe this is ready, pending your review.

I have no idea what the codecov/project check is complaining about. :)

drivers/storage/vbox/storage/vbox_storage.go
```diff
@@ -674,7 +675,7 @@ func (d *driver) attachVolume(
 		return err
 	}
 
-	if err := m.Refresh(); err != nil {
+	if err = m.Refresh(); err != nil {
```
akutz, line 678: Hi @codenrhoden,

I have a question. Here, and the other locations in this file, why did you switch to re-using the previously declared `err` var? I see no reason to do so, and keeping  it local to the scope of the if blocks keep it scoped to a specific usage, no?

codenrhoden: I did it to make the linter happy. I can revert that.

akutz: Hi @codenrhoden,

Ah, getting the shadow warning? You're fine. Just curious.

codenrhoden: Yes it was the shadow warning. It's still not clear to me what the right thing to do there is, as you make a good point about scoping.

akutz: Hi @codenrhoden,

While I agree about scoping, the linter is concerned with [shadowing](https://golang.org/doc/go1#shadowing) -- the unintentional process of referring to a different variable of the same name.

For example:

```go
func (r *router) volumes(
	ctx types.Context,
	w http.ResponseWriter,
	req *http.Request,
	store types.Store) error {

	filter, err := parseFilter(store)
	if err != nil {
		return err
	}
	if filter != nil {
		store.Set("filter", filter)
	}

	var (
		tasks   = map[string]*types.Task{}
		taskIDs []int
		opts    = &types.VolumesOpts{
			Attachments: store.GetAttachments(),
			Opts:        store,
		}
		reply = types.ServiceVolumeMap{}
	)

	for service := range services.StorageServices(ctx) {

		run := func(
			ctx types.Context,

			ctx = context.WithStorageService(ctx, svc)

			var err error
			if ctx, err = context.WithStorageSession(ctx); err != nil {
				return nil, err
			}

			return getFilteredVolumes(ctx, req, store, svc, opts, filter)
		}

		task := service.TaskExecute(ctx, run, schema.VolumeMapSchema)
		taskIDs = append(taskIDs, task.ID)
		tasks[service.Name()] = task
	}
```

In the above code snippet from [`api/server/router/volume/volume_routes.go#L20-L63`](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L20-L63), the `ctx` variable in the lambda `run` shadows the `ctx` variable from the outer function `volumes`. I do this on purpose as there is no harm and I prefer the consistency of the same variable name. However, someone may, unaware, mean to refer to the `ctx` from the parent `volumes` function from inside the `run` lambda, and it will not work.

Issue #352, opened 2016-12-06
Attachment mask that requires instance ID or local devices should return HTTP 400 if that data is missing


akutz:
When an API call with an attachment mask that requires the instance ID or local device headers is made, the call should return an HTTP status code of 400 (invalid request) instead of 500 (internal server error) if the header information is missing. As reported by @cantbewong.

PR #494, opened 2017-03-29
Volume Requests w Missing IID/LocDevs ret 400

akutz:
This patch handles issue #352 so that API requests with an attachment mask that requires either an instance ID or local devices map will return HTTP status 400 (bad request) rather than HTTP status 500 (server-side error).

### DIFF
--- a/api/server/handlers/handlers_errors.go
+++ b/api/server/handlers/handlers_errors.go
@@ -52,6 +52,9 @@ func getStatus(err error) int {
 		return http.StatusUnauthorized
 	case *types.ErrNotFound:
 		return http.StatusNotFound
+	case *types.ErrMissingInstanceID,
+		*types.ErrMissingLocalDevices:
+		return http.StatusBadRequest
 	default:
 		return http.StatusInternalServerError
 	}
--- a/api/types/types_errors.go
+++ b/api/types/types_errors.go
@@ -22,6 +22,10 @@ type ErrBadAdminToken struct{ goof.Goof }
 // resource that cannot be found.
 type ErrNotFound struct{ goof.Goof }
 
+// ErrMissingLocalDevices occurs when an operation requires local devices
+// and they're missing.
+type ErrMissingLocalDevices struct{ goof.Goof }
+
 // ErrMissingInstanceID occurs when an operation requires the instance ID for
 // the configured service to be avaialble.
 type ErrMissingInstanceID struct{ goof.Goof }
--- a/api/utils/utils_errors.go
+++ b/api/utils/utils_errors.go
@@ -38,6 +38,13 @@ func NewMissingInstanceIDError(service string) error {
 	}
 }
 
+// NewMissingLocalDevicesError returns a new ErrMissingLocalDevices error.
+func NewMissingLocalDevicesError(service string) error {
+	return &types.ErrMissingLocalDevices{
+		Goof: goof.WithField("service", service, "missing local devices"),
+	}
+}
+
 // NewStoreKeyErr returns a new ErrStoreKey error.
 func NewStoreKeyErr(key string) error {
 	return &types.ErrStoreKey{


codenrhoden:
HI @akutz.  I see this is already merged, but a couple of questions...

Is there more going into this to solve #352? I don't see any code that uses the new `ErrMissingLocalDevices` type.

I also wonder if it is appropriate to have all the drivers updated to use this as well now, as most of them do this type of checking, like [here](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/gcepd/storage/gce_storage.go#L750) and [here](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/azureud/storage/azure_storage.go#L642).

And one side note -- if you do want the merged PR to close an existing issue, the issue number should be prefixed with "closes" or "fixes".

akutz:
Hi @codenrhoden,

> And one side note -- if you do want the merged PR to close an existing issue, the issue number should be prefixed with "closes" or "fixes".

Thanks!

> Is there more going into this to solve #352? I don't see any code that uses the new `ErrMissingLocalDevices` type.

No, because people who report issues don't always consider the full scope of them. It's certainly a nice position in which to be :) It's not up to the framework to determine when and where local devices are needed or required. I provide the error and the handling of it, but it's up to the drivers to do the right thing. I also don't consider it our role to necessarily handle all the drivers as part of an issue scoped to the framework.

> I also wonder if it is appropriate to have all the drivers updated to use this as well now, as most of them do this type of checking, like [here](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/gcepd/storage/gce_storage.go#L750) and [here](https://github.com/codedellemc/libstorage/blob/master/drivers/storage/azureud/storage/azure_storage.go#L642).

Perhaps. I don't care? I mean to say that I'm concerned at the framework level, not the driver level. It's up to the drivers to return the correct error. 

The one thing the framework *can* and *does* do is determine if, based on an attachment mask, whether or not the instance ID is required. That's why I can reject requests before they ever get to a driver if an instance ID is missing and the mask indicates it is required.

However, it's also possible that drivers could need the instance ID in more than those occasions and thus should return these errors in those instances as well. I did not scrub through the drivers to find that out. I simply introduced handling these two types of errors with a `400` instead of a `500` per the issue. At the time I filed the issue on someone's behalf I did not realize there wasn't a way to centrally detect if the local devices map was required per the mask as drivers handle / require that data differently.

codenrhoden:
Cool. So what I'm getting from this is that yes, I can and should take advantage of those new error types and methods in storage drivers. Thanks! :)

akutz:
Hi @codenrhoden,

Yes, and everyone always should have as the instance ID error has been around a while. The other thing to discern is that it's not really possible for the framework to handle a missing local devices header centrally whereas there are several cases where it is possible to do the same for the instance ID. That's why when I see drivers *not* using the `MustInstanceID` variant I will correct them as, for example, there's no way a driver would receive a message to attach a volume if there is no instance ID present.

codenrhoden:
You are not wrong... and as I get better at this there will be a second pair of eyes to help make things consistent. Keep in mind that no driver (other than VFS, as you pointed out, and EFS, albeit incorrectly) used the `types.ErrNotFound` error either, until #485.

I understand what you are saying about `MustInstanceID`. That's a good pointer.

I will argue one point though. I think there is a scenario where the framework *can* detect and enforce the requirement for a local devices header. We have stated that the local devices header is required whenever the attachments flag has the devices bit set: https://github.com/codedellemc/libstorage/blob/master/api/types/types_drivers_storage.go#L188-L194

It is true that not all drivers need it, but I still think it is fair that if the the devices bit is set, and the local devices header is absent, the framework should return an error. What do you think? I'm happy to submit a PR for that if you agree.

akutz:
Hi @codenrhoden,

You got me on a technicality. The comment does tie this to the header, but I would actually change the comment and instead state that the bit indicates a mapping should occur and that a driver __could__ use that header info. For now I'm fine with what you proposed, but in the future I'm going to start refactoring things to avoid mentioning things like *header* in the API code.

codenrhoden:
Well I don't want to add it if it doesn't add value. I'm not trying to get you or trick you into it. :) I can say that from what understanding I had, I did consider that header required to do local device look up, even if a driver doesn't use it. I guess it's a question of whether we should enforce that. I'm okay either way, let's pick whichever we think is the better solution.

PR #495, opened 2017-03-29
Add minimum vol size, reject request too small

### LINKED ISSUE #459: Ensure all drivers have a valid default size and reject size 0
@vladimirvivien commented on [Thu Aug 18 2016](https://github.com/codedellemc/rexray/issues/536)

Currently the following will work (but backend will reject request).

```
$>rexray volume create --volumename="somename"
```

Most storage backend require a size when creating a volume, the CLI probably should do the same.

---

@akutz commented on [Thu Aug 18 2016](https://github.com/codedellemc/rexray/issues/536#issuecomment-240771853)

Hi @vladimirvivien,

I would normally agree except REX-Ray and libStorage both understand and support the concept of a default size. That's why the lack of the flag is not rejected. 

---

@vladimirvivien commented on [Thu Aug 18 2016](https://github.com/codedellemc/rexray/issues/536#issuecomment-240814601)

@akutz 
I thought that was the case as well, but I just redid this test against the latest rexray without the size param and it failed.  This is using the scaleio driver.  The driver got size=0.  Should this be a bug against libStorage ?

```
$> rexray volume create --volumename="vol-test-rexray" -l debug
...
RRO[0001] error creating volume                         IOPS=0 availabilityZone= host=unix:///var/run/libstorage/282555454.sock inner=problem getting response: Wrong command parameters. You can check that you are using the correct parameters, by consulting the help for the command. instanceID=scaleio=0720519500000000 provider=scaleIO route=volumeCreate server=puzzle-pixie-sr service=scaleio size=0 task=0 time=1471482712815 tls=false txCR=1471482712 txID=995a32b1-9ec2-41c2-708d-a461f2b3ec2a volumeName=vol-test-rexray volumeType=default
```

---

@akutz commented on [Thu Aug 18 2016](https://github.com/codedellemc/rexray/issues/536#issuecomment-240818136)

I don't know. Please re-execute using all available logging. Use the tip of REX-Ray and execute the command using the embedded mode with `REXRAY_DEBUG=true` prefixed on the command line. This enables all available debug logging. It will show what is being sent and consumed between the libStorage client and server.

The config value [_is_ registered](https://github.com/emccode/libstorage/blob/master/drivers/integration/docker/docker.go#L588) with a default value of 16. I didn't write this code, and I want to refactor it anyway so that it is no longer registered as some randomly-GB value. We should support size suffixes to indicate MB or GB, etc. 

My initial thought is the bug could be in REX-Ray. If the `size` flag is not provided, the value parsed from it is 0.

---

@akutz commented on [Thu Aug 18 2016](https://github.com/codedellemc/rexray/issues/536#issuecomment-240818810)

And that's what is happening. The `size` [field](https://github.com/emccode/rexray/blob/master/rexray/cli/cli.go#L98) in REX-Ray is an `int64`, so if the value is not provided it will always be zero. We _could_ make this a pointer to an integer so it can be nil. However, I didn't write any of this code. The config value should _not_ be a string when this value is an integer. We should make the CLI value a _string_ and parse it as a size with allowed suffixes. If the string is empty, the CLI flag not set, we simply don't pass the value to libStorage instead of providing a value of 0 as we are doing now.

---

@codenrhoden commented on [Mon Feb 27 2017](https://github.com/codedellemc/rexray/issues/536#issuecomment-282882473)

Would a reasonable approach for this be to make sure that each storage driver has a default size value, built into the driver (and overridable via config).  The `linux` integration driver has a default size of 16GB, but if you don't go through the integration driver, a size 0 volume request can appear at the driver.  Some drivers reject this outright, some drivers defer to a default size, and yet still more (ahem, `rbd` gladly pass this to the backend storage).

I think having a per-driver default size might be a good idea. Several providers have minimums that must be met as well.

---

@akutz commented on [Mon Feb 27 2017](https://github.com/codedellemc/rexray/issues/536#issuecomment-282887241)

Hi @codenrhoden,

Hmmm. Maybe... Remember that REX-Ray is a client, libStorage is the API. The client, IMO, should not be important to the API to this extent. The integration driver is also client side an not at all required. For example, the client @vladimirvivien created for K8 doesn't use it.

I like your suggestion, but here's a spin on it. Just as the integration driver has defaults, perhaps what we need to introduce are defaults for the API's request parameters that a driver can elect to also override. For example:

```yaml
libstorage:
  server:
    api:
      volumes:
        create: 
          size: 10GB
  services:
    ebs-00:
      driver: ebs
    ebs-01:
      driver: ebs
      libstorage:
        server:
          api:
            volumes:
              create:
                size: 15GB
```

When the driver code requests the property `libstorage.server.api.volumes.create.size` when responding to a request issued to the `ebs-00` service, the result will be `10GB`. However, if a request is sent to the `ebs-01`, the property `libstorage.server.api.volumes.create.size` has a value of `15GB`.

These values would be defaults and only used if a request omits a property. I imagine this would be implemented at the server's router level instead of by individual drivers since, if implemented at the router level, the drivers don't have to duplicate code. It would be fairly trivial to do so in [`volume_routes.go`](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L380-L403), with some changes similar to the following:

```go
func (r *router) volumeCreate(
	ctx types.Context,
	w http.ResponseWriter,
	req *http.Request,
	store types.Store) error {

	service := context.MustService(ctx)

	run := func(
		ctx types.Context,

		opts := &types.VolumeCreateOpts{
			AvailabilityZone: store.GetStringPtr("availabilityZone"),
			IOPS:             store.GetInt64Ptr("iops"),
			Size:             store.GetInt64Ptr("size"),
			Type:             store.GetStringPtr("type"),
			Encrypted:        store.GetBoolPtr("encrypted"),
			EncryptionKey:    store.GetStringPtr("encryptionKey"),
			Opts:             store,
		}

		serviceConfig := r.config.Scope(
			fmt.Sprintf("libstorage.server.services.%s", service.Name())

		if opts.Size == nil {
			i := serviceConfig.GetInt(
				"libstorage.server.api.volumes.create.size")
			opts.Size = &i
		}

		v, err := svc.Driver().VolumeCreate(ctx, store.GetString("name"), opts)
```

codenrhoden:
For each driver that uses the opt.Size field to create a volume with the
requested size, set a minimum volume size. If the incoming request does
not set a size, default to that minimum size. If the incoming request
does set a size, make sure that it is >= to the minimum size. If it is
not, return an error.

This touches all drivers except those that are not block devices -- Isilon, S3FS, EFS.

Fixes #459

### DIFF
--- a/drivers/storage/scaleio/storage/scaleio_storage.go
+++ b/drivers/storage/scaleio/storage/scaleio_storage.go
@@ -21,7 +21,8 @@ import (
 )
 
 const (
-	cc = 31
+	cc         = 31
+	minSizeGiB = 1
 )
 
 type driver struct {
@@ -365,11 +366,21 @@ func (d *driver) VolumeCreate(ctx types.Context, volumeName string,
 	if opts.Type != nil {
 		volume.Type = *opts.Type
 	}
-	if opts.Size != nil {
-		volume.Size = *opts.Size
-	}
 	if opts.IOPS != nil {
 		volume.IOPS = *opts.IOPS
+	}
+
+	if opts.Size == nil {
+		volume.Size = int64(minSizeGiB)
+	} else {
+		volume.Size = *opts.Size
+	}
+
+	fields["size"] = volume.Size
+
+	if volume.Size < minSizeGiB {
+		fields["minSize"] = minSizeGiB
+		return nil, goof.WithFields(fields, "volume size too small")
 	}
 
 	vol, err := d.createVolume(ctx, volumeName, volume)
--- a/drivers/storage/azureud/storage/azure_storage.go
+++ b/drivers/storage/azureud/storage/azure_storage.go
@@ -46,6 +46,8 @@ const (
 
 	// Default new disk size
 	defaultNewDiskSizeGB int32 = 128
+
+	minSizeGiB = 1
 )
 
 type driver struct {
@@ -347,14 +349,14 @@ func (d *driver) VolumeCreate(ctx types.Context, volumeName string,
 	}
 
 	size := int64(defaultNewDiskSizeGB)
-	if opts.Size != nil && *opts.Size != 0 {
+	if opts.Size != nil && *opts.Size >= minSizeGiB {
 		size = *opts.Size
 	}
 	size *= size1GB
 
 	fields := map[string]interface{}{
-		"volumeName":    volumeName,
-		"size_in_bytes": size,
+		"volumeName":  volumeName,
+		"sizeInBytes": size,
 	}
 
 	blobClient := mustSession(ctx).blobClient
--- a/drivers/storage/dobs/storage/dobs_storage.go
+++ b/drivers/storage/dobs/storage/dobs_storage.go
@@ -19,6 +19,10 @@ import (
 	doUtils "github.com/codedellemc/libstorage/drivers/storage/dobs/utils"
 )
 
+const (
+	minSizeGiB = 1
+)
+
 type driver struct {
 	name   string
 	config gofig.Config
@@ -136,6 +140,10 @@ func (d *driver) VolumeCreate(
 	name string,
 	opts *types.VolumeCreateOpts) (*types.Volume, error) {
 
+	fields := map[string]interface{}{
+		"volumeName": name,
+	}
+
 	if opts.AvailabilityZone == nil || *opts.AvailabilityZone == "" {
 		instance, err := d.InstanceInspect(ctx, nil)
 		if err != nil {
@@ -143,6 +151,19 @@ func (d *driver) VolumeCreate(
 		}
 		opts.AvailabilityZone = &instance.Region
 	}
+
+	if opts.Size == nil {
+		size := int64(minSizeGiB)
+		opts.Size = &size
+	}
+
+	fields["size"] = *opts.Size
+
+	if *opts.Size < minSizeGiB {
+		fields["minSize"] = minSizeGiB
+		return nil, goof.WithFields(fields, "volume size too small")
+	}
+
 	volumeReq := &godo.VolumeCreateRequest{
 		Region:        *opts.AvailabilityZone,
 		Name:          name,
--- a/drivers/storage/ebs/storage/ebs_storage.go
+++ b/drivers/storage/ebs/storage/ebs_storage.go
@@ -37,6 +37,8 @@ const (
 	waitVolumeAttach = "attach"
 	// waitVolumeDetach signifies to wait for volume detachment to complete
 	waitVolumeDetach = "detach"
+
+	minSizeGiB = 1
 )
 
 type driver struct {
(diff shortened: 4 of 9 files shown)


codenrhoden:
@akutz This is rebased and ready for your review/merge after CI tests.

If it looks good, I'd like to get it in tonight or tomorrow morning, solely because of the work I'm doing in #500. Tomorrow I am going to build all those drivers' tests and run them on the cloud providers, so it'd be great to have the best chance of finding anything new from this or from #500.

drivers/storage/scaleio/storage/scaleio_storage.go
```diff
@@ -365,13 +366,21 @@ func (d *driver) VolumeCreate(ctx types.Context, volumeName string,
...
-	}
 	if opts.IOPS != nil {
 		volume.IOPS = *opts.IOPS
 	}
 
+	if opts.Size == nil {
+		volume.Size = int64(minSizeGiB)
+	}
+
+	fields["size"] = *opts.Size
```
akutz: Hi @codenrhoden,

This will result in a runtime panic if `opts.Size` is `nil` since you cannot take the address of a `nil` value. Your pattern in the other sources accounts for this:

```go
if opts.Size == nil {
	opts.Size = &minSizeGiB
}
```

akutz (changes requested):
Hi @codenrhoden,

Other than the bug I noted the only comment I have is why you elected to do this per driver and not via the framework?

codenrhoden:
@akutz Thanks for catching that bug!

I had a few reasons for doing it per-driver. Let's see if they make sense.

- I originally thought the drivers would vary more. Each one would have a different storage-provider specific value. In the end, they almost all end up being 1G, and GCE is 10G.
- Not all the drivers need this -- only the block storage drivers. I suppose if done at the framework, we can still check the driver `Type` and only do it for those...
- I was afraid that if done at the framework layer, it might lead to unwanted overrides. What I mean is that if the incoming value is `nil`, we know it wasn't set and so we defer to a reasonable default. If the size *is* set, we enforce that it meets a minimum. If it's below the minimum, we return an error. But if the framework always makes sure this value is set, we no longer know if the requested size was set explicitly or by the framework. So really it comes down to what to do for drivers that have a minimum size other than 1G? If a user requests 5G today on GCE, we return an error. If instead, we automatically created a 10G volume, because that's the minimum, I think it's reasonable to handle at the framework level.
- It was just a natural evolution from the starting point of having to check if the size is nil. If we set it at the framework level, it should never be nil and we could remove that check from the drivers.

codenrhoden:
Really glad you caught that issue in ScaleIO, @akutz. As i was fixing that, I noticed that if `*opts.Size` was *not* nil, I never assigned it to `volume.Size`. So the requested size would have been ignored... Doh.

ScaleIO is the only driver that takes all of the `opts` fields and copies them into a new struct, rather than using `opts.Size` directly. So that was why I missed it.

akutz (approved):

PR #498, opened 2017-04-03
update pflag version

codenrhoden:
I noticed when building REX-Ray from current master that glide was spitting out warnings because of competing required versions of `pflag`. This is because both projects have `pflag` in their `glide.yaml`, and REX's was recently updated. Update libStorage to match.

I also noticed there is no code in `libStorage` that does `import github.com/spf13/viper`, so I removed it from the `glide.yaml` file. It is a transitive dependency, and the version doesn't change when I re-ran `glide up`. If we need it locked down, I will remove it from this patch.


akutz (approved):

akutz:
Hi @codenrhoden,

You need to rebase this due to `glide.lock` having changed, but other than that I approved the PR.

PR #499, opened 2017-04-03
Rename TaskExecute to TaskEnqueue for clarity

codenrhoden:
Calling TaskExecute() is really putting a task on a queue, not executing
that task immediately/directly. Clarify that by renaming the function.


akutz (approved):
Well that makes way too much sense, doesn't it? :)

codenrhoden:
Ha. It was your idea. Well, we both said as much when you reviewed the task system with me. I just decided to do it. :)

PR #500, opened 2017-04-03
make sure storage driver polling functions have timeouts

### LINKED ISSUE #480: endless loop possible checking volume states
# Summary
When the EBS driver does an attach, create, or delete of a volume it calls `waitVolumeComplete`, which enters a loop polling the status of the volume once per second. If the desired state never appears, the loop never exits.

We need a hard timeout within the driver where we abort the check.

Reported via codedellemc/rexray#773

## Expected Behavior
If the expected state of a volume never shows up, the driver aborts and returns an error after a configurable timeout

## Actual Behavior
The driver never returns, which prevents all future EBS queries from executing.

## Steps To Reproduce
I have been unable to reproduce myself, but the issue is apparent via code inspection. Details again are in the previously linked Issue report in REX-Ray. Reproduction steps will hopefully be figured out to verify a fix.

codenrhoden:
Introduce a new `WaitFor` API utility that takes a lambda function and a
timeout. The function will call the lambda and return its results, or
the timeout will expire first and callers can check for this.

For each driver that was entering a loop to poll for some kind of volume
status, make sure they use this `WaitFor` utility so that the polling
loop has a hard timeout associated with it so it does not block the
driver indefinitely. It is also important that the polling loop itself
has a hard limit on the number of times it will run.

Since the polling loop (lambda) will be run in a goroutine, even if the
timeout is reached, the polling loop will still be running in the
background.

Fixes #480

Tests:

- [x] DigitalOcean
- [x] EBS
- [x] EFS
- [x] GCE

### DIFF
--- a/.docs/user-guide/storage-providers.md
+++ b/.docs/user-guide/storage-providers.md
@@ -49,7 +49,11 @@ ebs:
   accessKey:      XXXXXXXXXX
   secretKey:      XXXXXXXXXX
   region:         us-east-1
+  maxRetries:     10
   kmsKeyID:       arn:aws:kms:us-east-1:012345678910:key/abcd1234-a123-456a-a12b-a123b4cd56ef
+  statusMaxAttempts:  10
+  statusInitialDelay: 100ms
+  statusTimeout:      2m
 ```
 
 ##### Configuration Notes
@@ -61,8 +65,20 @@ access credentials, like environment variables or instance profile IAM permissio
 See official AWS documentation for list of supported regions.
 <!-- - `tag` is used to partition multiple services within single AWS account and is
 used as prefix for EBS names in format `[tagprefix]/volumeName`. -->
+- `maxRetries` is the number of retries that will be made for failed operations
+  by the AWS SDK.
 - If the `kmsKeyID` field is specified it will be used as the encryption key for
 all volumes that are created with a truthy encryption request field.
+- `statusMaxAttempts` is the number of times the status of a volume will be
+  queried before giving up when waiting on a status change
+- `statusInitialDelay` specifies a time duration used to wait when polling
+  volume status. This duration is used in exponential backoff, such that the
+  first wait will be for this duration, the second for 2x, the third for 4x,
+  etc. The units of the duration must be given (e.g. "100ms" or "1s").
+- `statusTimeout` is a maximum length of time that polling for volume status can
+  occur. This serves as a backstop against a stuck request of malfunctioning API
+  that never returns.
+
 
 For information on the equivalent environment variable and CLI flag names
 please see the section on how non top-level configuration properties are
@@ -153,6 +169,9 @@ efs:
   region:              us-east-1
   tag:                 test
   disableSessionCache: false
+  statusMaxAttempts:  6
+  statusInitialDelay: 1s
+  statusTimeout:      2m
 ```
 
 ##### Configuration Notes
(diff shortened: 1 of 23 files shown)


codenrhoden:
@akutz I wanted to put this up for early review, so that you can see what I was thinking with it. I started with DigitalOcean for a driver.

One thing I noted with the DOBS driver is that it had a form of exponential backoff on its retries when looking for a specific state. This is something I would like to carry over to other drivers, unless you object.

codenrhoden:
I've implemented the core of what I wanted to do here. Next I need to test each driver. I will try building the unit tests for each driver as well, just to make sure that those still work.

Not sure if I will be able to test fittedcloud or not. But the changes I made to that driver are identical to the changes for EBS, so I'm confident that anything wrong would be turned up in EBS testing.

.docs/user-guide/storage-providers.md
```diff
@@ -217,19 +247,20 @@ Below is a working `config.yml` file that works with AWS EFS.
...
-        efs:
-          accessKey:      XXXXXXXXXX
-          secretKey:      XXXXXXXXXX
-          securityGroups:
-          - sg-XXXXXXX
-          - sg-XXXXXX0
-          - sg-XXXXXX1
-          region:         us-east-1
-          tag:            test
+efs:
```

I appreciate you fixing the doc examples, but I also want to isolate PRs to a specific purpose. Let's leave this change, but just FYI for the future.

.docs/user-guide/storage-providers.md
```diff
@@ -675,22 +706,58 @@ The DigitalOcean block storage driver has the following requirements:
 * Valid DigitalOcean [access token](https://goo.gl/iKoAec)
 
 #### Configuration
-The following example illustrates a minimal configuration for the DigitalOcean
+The following is an example with all possible fields configured. For a running
+example see the `Examples` section.
```
akutz, line 708: Hi @codenrhoden,

> see the `Examples` section.

I always like to add a _please_ in front of a directive. Just a nice thing to do :) Also, if you're asking someone to look at something why not link to it with an anchor? It's also possible to add custom anchors to avoid using the ones generated by mkdocs. Here's an example:

```
<a class="headerlink hiddenanchor" name="aws-ebs"></a>
```

.docs/user-guide/storage-providers.md
```diff
@@ -675,22 +706,58 @@ The DigitalOcean block storage driver has the following requirements:
...
+  queried before giving up when waiting on a status change
+- `statusInitialDelay` specifies a time duration used to wait when polling
+  volume status. This duration is used in exponential backoff, such that the
+  first wait will be for this duration, the second for 2x, the third for 3x,
+  etc. The units of the duration must be given (e.g. "100ms" or "1s").
+- `statusTimeout` is a maximum length of time that polling for volume status can
+  occur. This serves as a backstop against a stuck request of malfunctioning API
+  that never returns.
+
+!!! note
```

I prefer not to have two note-styled notes back-to-back. Let's either combine them or separate them somehow.

.docs/user-guide/storage-providers.md
akutz: Eh, nevermind this. I thought this was a DigitalOcean PR when I wrote this comment. Stupid me.

codenrhoden:
@akutz FWIW, I didn't think your comment on changing the EFS config template was stupid at all. It's not related to what I was changing here, and therefore should probably not be a part of this. I am usually a stickler for that as well. I will probably remove those changes from this PR, and do a separate one that is only doc focused.

I saw that because I was also introducing a new change, which is to add a `libstorage.service` field to the example, to help reduce the number of times users get tripped up on that. But I only added that to drivers I was modifying... so I would need to do a second PR to pick up the others anyways. I should separate that improvement out.

.docs/user-guide/storage-providers.md
codenrhoden: Agreed. I'll take that enhancement on in a separate doc PR. The majority of the providers have this line [verbatim](http://libstorage.readthedocs.io/en/stable/user-guide/storage-providers/#configuration_8), and none of them have a link.

It's always just the case of taking precedent from an existing driver, and using that as a guide for the next one. :)

codenrhoden:
Okay, this is in a working state for final review.

The scope grew once I started to try to compile and and run the driver unit tests. Only GCE didn't need changes.

For EBS, the IID tests would fail when running in "compatibility" mode as the `ec2` driver because the IID had a hardcoded `ebs` as the Driver name. I also hit a real error where the EBS driver would crash with a nil dereference if volume create request had a nil `opts.AvailabilityZone`. Surprised that hasn't come up before.

The EFS driver had an issue with a nil context, but that was an easy fix.

Digital Ocean tests had never been updated after the driver was renamed from `digitalocean` to `dobs`.

codenrhoden:
@akutz can you give this another look today?

PR #501, opened 2017-04-05
libStorage Token Support (JWT)

akutz:
This patch introduces support for JSON Web Tokens, used to restrict access to API resources to clients that provide valid tokens with sufficient claims. Please see the commit's doc section for additional information. The docs are also [built](http://libstorage-akutz.readthedocs.io/en/feature-token-support/user-guide/config/#authentication) on my fork at RTFD.

### DIFF
--- a/.docs/user-guide/config.md
+++ b/.docs/user-guide/config.md
@@ -451,6 +451,204 @@ that has signed the allowed client certificates. Or at the very least the
 server certificate would be signed by the same intermediate CA that is used
 to sign the client-side certs.
 
+### Authentication
+In addition to TLS, the libStorage API includes support for
+[JSON Web Tokens](https://jwt.io) (JWT) in order to provide authentication
+and authorization.
+
+A JWT is transmitted along with an API call in the standard HTTP `Authorization`
+header as an OAuth 2.0 [Bearer](https://tools.ietf.org/html/rfc6750) token. For
+example:
+
+```
+GET /volumes
+
+Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE1MjI2ODg1NTAsImlhdCI6MTQ5MTIzODk1MCwibmJmIjoxNDkxMjM4OTUwLCJzdWIiOiJha3V0eiJ9.3eAA7AQZUGrwA42H64qKbu8QF_AHpSsJSMR0FALnKj8
+```
+
+The above token is split into three discreet parts:
+
+`HEADER`.`PAYLOAD`.`SIGNATURE`
+
+The decoded header is:
+```json
+{
+  "alg": "HS256",
+  "typ": "JWT"
+}
+```
+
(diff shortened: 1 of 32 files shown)


.docs/user-guide/config.md
```diff
@@ -451,6 +451,196 @@ that has signed the allowed client certificates. Or at the very least the
...
+  server:
+    auth:
+      key: MySuperSecretSigningKey
+      alg: HS256
+      allow:
+      - akutz
+```
+
+The above configuration snippet defines a new property,
+`libstorage.server.auth`. This property contains three child properties:
```
codenrhoden: s/three/five/

api/server/security/security_token.go
```diff
@@ -0,0 +1,151 @@
...
+	lf["sub"] = tok.Subject
+	lf["iat"] = tok.IssuedAt
+	lf["exp"] = tok.Expires
+	lf["nbf"] = tok.NotBefore
+
+	var (
+		allowed bool
+		denied  bool
+	)
+	for _, v := range config.Deny {
```
codenrhoden: This code that loops through Denied and Allow subjects identical to that found in the service auth handler (when a previously validated token is already in the context). Could we move this bit of logic in a common function, such as `security.ValidateTokenSubject(config, tok.Subject)` and call that from both places?

api/types/types_errors.go
```diff
@@ -51,3 +53,34 @@ type ErrBatchProcess struct{ goof.Goof }
 // ErrBadFilter occurs when a bad filter is supplied via the filter query
 // string.
 type ErrBadFilter struct{ goof.Goof }
+
+// ErrMissingStorageService occurs when the storage service is expected in
+// the provided context but is not there.
+var ErrMissingStorageService = errors.New("missing storage service")
```
codenrhoden: Why the use of `errors` instead of `goof` here? This one is defined differently than `ErroTimedOut` and `ErrNotImplemented`

codenrhoden (changes requested):
This is looking good. Apologies that two of my review comments went out as single comments rather than all at once in single review!

I know you noted this already, but just for posterity there is not any logic in place that would return an HTTP 401 error if you do an API call on multiple services w/o the JWT and one of the services requires it.

api/types/types_errors.go
akutz: Hi @codenrhoden,

This has been handled.

akutz:
Hi @codenrhoden,

FYI there *is* now logic to handle the multi-service auth. I created a new handler at [`handlers_auth_all_services.go`](https://github.com/akutz/libstorage/blob/feature/token-support/api/server/handlers/handlers_auth_all_services.go) that is applied to the following routes:

* `/services/`
* `/volumes/`
* `/snapshots/`

The latest update also handles the duplicate code we discussed. Thank you again for providing such a comprehensive review.

codenrhoden (approved):

PR #506, opened 2017-04-11
Fix Logging on Volume Creation

### LINKED ISSUE #478: Fix logging of EBS volume creation
# Summary
When creating an EBS volume the line that logs the volume's details logs the addresses of a volume struct's fields instead of the values.

Many thanks to @meatlewis for discovering this issue!

## Expected Behavior
A line that looks like this:

```bash
time="2017-03-21T16:21:45Z" level=debug msg="creating volume" driverName=ebs opts=&{0xc4207419f0 0xc420741a00 0xc420741a10 0xc420741a20 <nil> <nil> map[opts:map[encrypted:true size:5 volumetype:gp2] service:ebs name:dvdcli iops:375 size:125 type:"gp2" availabilityzone:"us-east-1a" encrypted:<nil> encryptionkey:<nil>]} volumeName=dvdcli 
```

## Actual Behavior
A line that looks like this:

```bash
time="2017-03-21T16:21:45Z" level=debug msg="creating volume" driverName=ebs opts=&{0xc4207419f0 0xc420741a00 0xc420741a10 0xc420741a20 <nil> <nil> map[opts:map[encrypted:true size:5 volumetype:gp2] service:ebs name:dvdcli iops:0xc420741a00 size:0xc420741a10 type:0xc420741a20 availabilityzone:0xc4207419f0 encrypted:<nil> encryptionkey:<nil>]} volumeName=dvdcli 
```

## Steps To Reproduce
Create an EBS volume.

## Logs
It is very important when filing an issue to include associated logs. There are
two different logs about which to be concerned: the service log (if libStorage
is running as a service) and the client log.

### Service Log
An example of the service log can be found [here](https://gist.github.com/akutz/3306701b4eb3de4dbf37efd5c80088d8#file-encrypted-ebs-vol-log-sh-L499).

akutz:
This patch handles fixes #478 by centralizing logging of volume creation activity into the volume route source away from the driver implementation. The error state is also handled in the route source meaning drivers should no longer log the error directly in their implementation of `VolumeCreate`.

### DIFF
--- a/api/server/router/volume/volume_routes.go
+++ b/api/server/router/volume/volume_routes.go
@@ -389,26 +389,51 @@ func (r *router) volumeCreate(
 		ctx types.Context,
 		svc types.StorageService) (interface{}, error) {
 
-		v, err := svc.Driver().VolumeCreate(
-			ctx,
-			store.GetString("name"),
-			&types.VolumeCreateOpts{
-				AvailabilityZone: store.GetStringPtr("availabilityZone"),
-				IOPS:             store.GetInt64Ptr("iops"),
-				Size:             store.GetInt64Ptr("size"),
-				Type:             store.GetStringPtr("type"),
-				Encrypted:        store.GetBoolPtr("encrypted"),
-				EncryptionKey:    store.GetStringPtr("encryptionKey"),
-				Opts:             store,
-			})
+		volumeName := store.GetString("name")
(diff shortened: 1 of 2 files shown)


codenrhoden (approved):
Very much :+1: Nice improvement.

And you fixed the typo while I was reviewing it. :)

PR #509, opened 2017-04-12
Path Enhancements

akutz:
This patch enhances the way libStorage manages its internal file path workflow.

* There is a new path constant, `TLS`. It's default path is `/etc/libstorage/tls`.

* When `LIBSTORAGE_HOME` is set the path structure removes the nested `libstorage` element. For example, while the default configuration directory is `/etc/libstorage`, now if `LIBSTORAGE_HOME` is set to `/tmp/libstorage` the new configuration directory path will be `/tmp/libstorage/etc` instead of `/tmp/libstorage/etc/libstorage`.

* All path constants can now be explicitly specified via environment variables, independent of the value of `LIBSTORAGE_HOME`. For example, if `LIBSTORAGE_HOME` is set to `/tmp/libstorage`, `LIBSTORAGE_PATHS_ETC` can be set to `/etc/libstorage` to ensure that the configuration directory is not nested inside of `LIBSTORAGE_HOME`. Available environment variables are:

    * `LIBSTORAGE_PATHS_ETC`
    * `LIBSTORAGE_PATHS_LIB`
    * `LIBSTORAGE_PATHS_LOG`
    * `LIBSTORAGE_PATHS_RUN`
    * `LIBSTORAGE_PATHS_TLS`
    * `LIBSTORAGE_PATHS_LSX`

### DIFF
--- a/api/types/types_paths.go
+++ b/api/types/types_paths.go
@@ -15,6 +15,69 @@ import (
 	"github.com/akutz/gotil"
 )
 
+var (
+	libstorageHome = os.Getenv("LIBSTORAGE_HOME")
+	fileKeyCache   = map[fileKey]string{}
+	homeLock       = &sync.Mutex{}
+	thisExeDir     string
+	thisExeName    string
+	thisExeAbsPath string
+	slashRX        = regexp.MustCompile(`^((?:/)|(?:[a-zA-Z]\:\\?))?$`)
+
+	etcEnvVarPath string
+	libEnvVarPath string
+	logEnvVarPath string
+	runEnvVarPath string
+	tlsEnvVarPath string
+	lsxEnvVarPath string
+)
+
+func init() {
+
+	thisExeDir, thisExeName, thisExeAbsPath = gotil.GetThisPathParts()
+
+	if libstorageHome == "" {
+		libstorageHome = "/"
+	}
+
+	// if not root and home is /, change home to user's home dir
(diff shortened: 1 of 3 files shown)


api/types/types_paths.go
```diff
@@ -121,43 +191,88 @@ func (k fileKey) key() string {
...
+			return "/var/log"
+		}
 		return "/var/log/libstorage"
 	case Run:
+		if isLibstorageHomeSet() {
+			return "/var/run"
+		}
 		return "/var/run/libstorage"
+	case TLS:
+		return "tls"
```
codenrhoden, line 234: Why is TLS different from the others, and not ever dependent on LIBSTORAGE_HOME ?

codenrhoden (approved):
Looks good. Just have the one question.

api/types/types_paths.go
akutz: Hi @codenrhoden,

Because of the function [`parent`](https://github.com/akutz/libstorage/blob/ac4fa0e8d091c51bedbd2bd166de191326cb4dd3/api/types/types_paths.go#L173):

```go
func (k fileKey) parent() fileKey {
	switch k {
	case TLS:
		return Etc
	case LSX:
		return Lib
	default:
		return Home
	}
}
```

Unless specified by an env var, the TLS path is relative to the Etc path.

codenrhoden: 👍

PR #510, opened 2017-04-14
Security Enhancements

akutz:
This patch enhances libStorage security:

* If discovered in `$LIBSTORAGE_HOME_ETC_TLS` the following files are automatically loaded:
    * `libstorage.crt`
    * `libstorage.key`
    * `cacerts`<br/><br/>

* If `$LIBSTORAGE_HOME_ETC_TLS/known_hosts` exists it is automatically loaded unless the property `libstorage.tls.knownHosts` is explicitly defined. This is the system's `known_hosts` file.

* If `$HOME/.libstorage/known_hosts` exists it is automatically used when TLS security is set to verify peer certificates. This is the user's `known_hosts` file.

* The above `known_hosts` files are line-delimited with each line following the format: `HOST ALGORITHM FINGERPRINT`

* When matching a remote host's known host information the peer's host (derived from the certificate's Subject.CommonName) is also taken into account. Thus if a host is marked as trusted but later provides a different certificate during TLS negotiation the connection will fail. This is identical to SSH's known host logic.

* The property `libstorage.tls.verifyPeers` is introduced. It's a boolean flag that indicates TLS connections should be verified against a known list of peer certificate fingerprints in the system's and user's `known_hosts` files.

  Enabling this property also sets `libstorage.tls.insecure` to `true`. The connection will be encrypted, but the certificate verification is disabled and deferred to the peer verification.

* The property `libstorage.tls` can now be set to a simple string value of `verifyPeers` to indicate TLS connections should be verified against the system's and user's `known_hosts` files.


akutz:
Hi @codenrhoden,

I accidentally pushed this branch to the upstream repo hence the forever-failed build (linked to that repo). The PR build is passing however.

PR #512, opened 2017-04-20
Paths Refactor

akutz:
This patch removes all of the libStorage path constants from the `github.com/codedellemc/libstorage/api/types` package and replaces package-level path management with a Context injected `*PathConfig` object from the `types` package.
    
The `github.com/codedellemc/libstorage/api/utils` package contains a new function named `NewPathConfig(ctx, home, token) *PathConfig`. This function lets developers specify a custom home directory for the path structure or even a custom token. A custom token will rebrand the elements of the path structure so they no longer reflect "libstorage" but the value specified by the token.
    
The new "paths" package also employs the Ginkgo natural-language test framework in combination with GΩmega to produce test coverage that's both more comprehensive as well as easier to read. Many thanks to @spiegela for recommending this approach!


api/paths/paths.go
```diff
@@ -1,80 +1,138 @@
-package types
+package paths
 
 import (
 	"bytes"
 	"fmt"
 	"io"
+	"log"
```
codenrhoden: Why the move away from contextual logging in this package with `logrus`? And also why remove `gotil`? Just trying to reduce dependencies?

api/paths/paths.go
```diff
@@ -1,80 +1,138 @@
...
 func init() {
+	monFunc := func(pEV, pV *string) {
+		if v := os.Getenv(*pEV); v != "" && v != *pV {
+			Init(appToken)
+		}
+	}
+	monLoop := func(pEV, pV *string) {
+		go func() {
+			for {
+				monFunc(pEV, pV)
```
codenrhoden: This is more of a golang question, but how is this not an efficiency killer, even with goroutines? Since there is no timer/delay, etc, it seems like this would create *something* that goes into a tight loop and pegs a core on your CPU or something.

api/paths/paths.go
```diff
@@ -1,80 +1,138 @@
...
+	monLoop(&envVarHomeLog, &appHomeLogPath)
+	monLoop(&envVarHomeRun, &appHomeRunPath)
+	monLoop(&envVarHomeLSX, &appHomeLSXPath)
+}
+
+const (
+	defaultAppToken = "libstorage"
+)
+
+var appToken = defaultAppToken
```
codenrhoden: Would you be opposed to moving this `const` and `var` definition above `init`? I like to see them all up at the top, but even more, when I read through `init` my first thought was, "but where was `appToken` defined" ?

api/paths/paths.go
```diff
@@ -203,7 +262,7 @@ func (k fileKey) perms() os.FileMode {
 func (k fileKey) key() string {
 	switch k {
 	case Home:
-		return "home"
+		return "hom"
```
codenrhoden: This change intended?

codenrhoden (commented):
This will be a really cool set of changes.

So, existing drivers will still work with their existing methods of registering config parameters? Is there anything they need to do?

api/paths/paths.go
akutz: Hi @codenrhoden,

For this package I specifically wanted to make sure it was *only* using stdlib due to the fact it will likely be the very first package anything invokes. I wanted to make sure it did not have dependencies on anything else.

api/paths/paths.go
akutz: Hi @codenrhoden,

That's a good point. My answer was going to be because when you range over a channel you don't have that issue. Then I realized I was no longer ranging over a channel because I had change the design. So yeah, fair point. I'll throw in a delay of a tick or something.

api/paths/paths.go
akutz: Hi @codenrhoden,

This is changed in a code update anyway, but yes, I agree.

api/paths/paths.go
akutz: Hi @codenrhoden,

Yes, if you read the tokens you'll note they're all three characters long. I wanted to fix the `home` token to keep it consistent.

akutz:
Hi @codenrhoden,

You may wish to take another look at this. I refactored it again. It's now far more manageable and predictable.

PR #517, opened 2017-04-25
Test Harness Fix

akutz:
This patch fixes the test harness so that it only registers configs once but creates a context every time that it's nil.


codenrhoden:
👍

PR #518, opened 2017-04-25
Enable Override of EBS Region, AZ, & Endpoint Props

akutz:
This patch introduces the ability for the EBS driver to override the client's provided region, availability zone, and endpoint properties by specifying them in the server-side config with a suffix of !important (like CSS).

Many thanks to libStorage & REX-Ray community member @shamimgeek for discovering this! Thanks Shamim!


codenrhoden (commented):
Looks good at first glance. Definitely want to hear it works from @shamimgeek. This will be a nice way to pick up eucalyptus support!

PR #519, opened 2017-04-26
ScaleIO, EBS, EFS, increased test coverage (codecov)

davidvonthenen:
This adds a mechanism for increased go test coverage for drivers. The high-level behavior is that we only want to run the driver tests on a single sub-build job in Travis CI. It made sense that the selected sub-build would be one in which if the driver tests failed, the build would fail which immediately ruled out using anything that was flagged in the .travis.yml as an allowed failure. All environment creation, testing, and environment teardown are done in parallel meaning that the time added to the build will always be the driver in which it takes the longest stand up and teardown. Based on my testing and observations, the original build time for that sub-build (usually [Build Number].5) took about 7 minutes to complete and with this change has increased by 5 minutes to about 12 minutes for the entire build to complete. Overall project code coverage has increased from ~20% to ~35% with this change.

The driver tests are enabled by the existing COVERAGE_ENABLED(=1), the existing environment variable DRIVERS(="scaleio ebs efs") to select which environment you want to add to the code coverage testing, an existing environment variable COVERED_GO_VERSION=(1.8) to select the specific go version to test on, new environment variables AWS_ACCESSKEY and AWS_SECRETKEY to manage driver test environments in AWS, and a new environment variable LAUNCH_KEY_NAME which is the specified AWS ssh key to use to access the AWS EC2 instances. These parameters can all be set in the environment variable section in Travis CI. **NOTE: the above environment variables have not been set until I have a chance to talk with stakeholders. In the meantime, this PR build will fail. Added a few new environment variables, but they aren't sticking. Maybe I dont have access.**

This PR leverages some existing functionality that already existed prior. Namely the concept of using 3 scripts already present in some of the driver test folders to create the environment (test-env-up.sh), to runs the tests (test-run.sh), and to teardown the environment (test-env-down.sh). 

Each drivers' test-env-up.sh script signified by the DRIVERS environment will be called in parallel at the very beginning of the build process. This will allow the environments to get created while the build is in progress and test binaries are being created.

A compiled go test binary for that particular driver will then be copied to the test environment via the test-run.sh script, the go coverage data will be created by the use of the -test.coverprofile flag, the raw go test data will be copied from the test environment and at the end, be submitted to codecov via APIs for analysis. 
 
The test-env-down.sh will clean up the environment and any potentially orphaned artifacts that the particular driver test *may* have created in the environment. In the normal "happy path" all artifacts will be cleaned up as a part of the driver test code. Should there be a build failure, the Makefile will short circuit and bomb out early thus skipping the invocation of these scripts. In this case, the Travis CI process will call the test-env-down.sh scripts instead as denoted in the .travis.yml.

### DIFF
--- a/drivers/storage/ebs/tests/ebs_test.go
+++ b/drivers/storage/ebs/tests/ebs_test.go
@@ -3,6 +3,7 @@
 package ebs
 
 import (
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -26,15 +27,34 @@ import (
 // Put contents of sample config.yml here
 var (
 	configYAMLec2 = []byte(`
+libstorage:
+  service: ec2
+  integration:
+    volume:
+      operations:
+        mount:
+          preempt: true
 ec2:
   region: us-west-2
-  tag: RR
-  endpoint: ec2.us-west-2.amazonaws.com`)
+  endpoint: ec2.us-west-2.amazonaws.com
+  accessKey: %s
+  secretKey: %s
+`)
+
 	configYAMLebs = []byte(`
+libstorage:
+  service: ebs
+  integration:
+    volume:
+      operations:
+        mount:
+          preempt: true
 ebs:
   region: us-west-2
-  tag: RR
-  endpoint: ec2.us-west-2.amazonaws.com`)
+  endpoint: ec2.us-west-2.amazonaws.com
+  accessKey: %s
+  secretKey: %s
+`)
 )
 
 var volumeName string
@@ -50,12 +70,28 @@ func skipTests() bool {
 
 // Set volume names to first part of UUID before the -
 func init() {
-	uuid, _ := types.NewUUID()
-	uuids := strings.Split(uuid.String(), "-")
-	volumeName = uuids[0]
-	uuid, _ = types.NewUUID()
-	uuids = strings.Split(uuid.String(), "-")
-	volumeName2 = uuids[0]
+	log.SetOutput(os.Stdout)
+	log.SetLevel(log.DebugLevel)
+
+	volumeName = os.Getenv("FIRST_VOLUME")
+	if len(volumeName) == 0 {
+		uuid, _ := types.NewUUID()
+		uuids := strings.Split(uuid.String(), "-")
+		volumeName = uuids[0]
+	}
+	volumeName2 = os.Getenv("SECOND_VOLUME")
+	if len(volumeName2) == 0 {
+		uuid, _ := types.NewUUID()
+		uuids := strings.Split(uuid.String(), "-")
+		volumeName2 = uuids[0]
+	}
+
+	// Build configuration based on provided environmet
+	awsAccessKey := os.Getenv("AWS_ACCESSKEY")
+	awsSecretKey := os.Getenv("AWS_SECRETKEY")
+
+	configYAMLec2 = []byte(fmt.Sprintf(string(configYAMLec2[:]), awsAccessKey, awsSecretKey))
+	configYAMLebs = []byte(fmt.Sprintf(string(configYAMLebs[:]), awsAccessKey, awsSecretKey))
 }
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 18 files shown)


drivers/storage/ebs/tests/ebs_test.go
```diff
@@ -138,7 +174,7 @@ func TestInstanceIDEC2(t *testing.T) {
 	}
 
 	// create storage driver
-	sd, err := registry.NewStorageDriver("ec2")
+	sd, err := registry.NewStorageDriver("ebs")
```
codenrhoden, line 177: Is this change necessary? I believe this part makes sure that things still work if someone is using `ec2` in their config file isntead of `ebs`. I think this test should pass without this change.

drivers/storage/ebs/tests/test-env-down.sh
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/env bash
+
+# This script cleans up the infrastructure used for tests
+
+#set -e
+
+# Make sure that jq is installed
```
codenrhoden, line 7: Comment is for `jq`, but package is `curl`

drivers/storage/ebs/tests/test-env-up.sh
```diff
@@ -0,0 +1,110 @@
...
+# that resources will be running.
+
+#set -e
+
+AWS_ACCESSKEY=${AWS_ACCESSKEY:-$1}
+AWS_SECRETKEY=${AWS_SECRETKEY:-$2}
+LAUNCH_KEY_NAME=${LAUNCH_KEY_NAME:-$3}
+CF_STACK_NAME=${CF_STACK_NAME:-$4}
+
+# Make sure that jq is installed
```
codenrhoden, line 14: s/curl/jq

drivers/storage/efs/tests/efs_test.go
```diff
@@ -22,56 +22,54 @@ import (
...
+        mount:
+          preempt: true
 efs:
-  tag: %s
-  region: %s
-  securityGroups: %s
+  region: us-west-2
+  endpoint: ec2.us-west-2.amazonaws.com
+  accessKey: %s
+  secretKey: %s
```
codenrhoden, line 41: Just as a general question across all drivers, and something I've been thinking about anyways, is how much of the config do we really want codified in these test files? It seems to me that as little as possible, and leaving it to environment variables is best. I would put in reasonable defaults here where it is helpful, but anything where someone *has* to make a change, I'd rather not see it here at all.  So rather than doing `os.getenv(AWS_ACCESSKEY)` down below, I'd rather see the tests run something like

`EFS_ACCESSKEY=$AWS_ACCESSKEY ./efs.test`

Thoughts?

drivers/storage/ebs/tests/test-env-down.sh
```diff
@@ -0,0 +1,109 @@
...
+hash curl 2>/dev/null || {
+  if [ -e "/etc/redhat-release" -o \
+         -e "/etc/redhat-version" ]; then
+    yum -y install curl
```
codenrhoden, line 11: In fact, why not just combine curl and jq in one install? will make the installs faster as well.

`yum -y install curl jq`
`apt-get install -y curl jq`
`brew install curl jq`

drivers/storage/scaleio/tests/scaleio_test.go
```diff
@@ -320,9 +334,9 @@ func TestVolumeAttach(t *testing.T) {
 		vol = volumeCreate(t, client, volumeName)
 		_ = volumeAttach(t, client, vol.ID)
 		_ = volumeInspectAttached(t, client, vol.ID)
-		_ = volumeInspectDetachedFail(t, client, vol.ID)
+		//_ = volumeInspectDetachedFail(t, client, vol.ID) TODO: has the behavior changed?
```
codenrhoden, line 337: Yes, it definitely did, but it's been a couple months. :)  The notion of when attachments should be returned very much depends on the query, and whether you've asked for attachments, attachmentsToMe, attachmentsWithDevices, etc.

here's an example of where I fixed this in EBS a couple weeks ago: https://github.com/codedellemc/libstorage/pull/500/files#diff-5812461cc888fb891da3e63c5d1433f0R214

drivers/storage/scaleio/tests/scaleio_test.go
```diff
@@ -25,14 +25,22 @@ import (
...
+  endpoint: https://10.0.0.11/api
   insecure: true
+  thinOrThick: ThinProvisioned
   userName: admin
-  password: Scaleio123
-  systemName: cluster1
-  protectionDomainName: pdomain
-  storagePoolName: pool1
+  password: F00barbaz
+  systemId: 0a79118e5ecc882c
```
codenrhoden, line 41: So where is set that these values are correct?  I don't see it in the CF template (but honestly don't know what I'm looking for). Again, i dislike seeing these, because I think it could trip people up who compile the test binary and choose to run it against their own ScaleIO cluster.

I'm not saying it's wrong though. All of these can (and likely should) be overriden by env vars. I'd just rather not compile in default credentials at all.

PR #526, opened 2017-04-30
Natural Language Testing

akutz:
This patch refactors the test package into an opaque, black box that utilizes Ginkgo testing in order to produce a comprehensive, natural language test harness.

This patch also enables building all of the tests by default.


codenrhoden (commented):
I like where this is going. I still have some reading up to do on ginkgo and gomega, but this is more what I've pictured as far as "refactor test harness" goes...  Natural language aside, I feel like we can have a standard suite of tests for drivers, that can be run against every driver easily. Most of what's in each driver's test package is a repeat across each one -- create a volume, query it a few ways, attach, detach, remove... All these things are the same for each one, and this looks like a way to standardize them all.

Is that a correct view of things?

It seems like if there is "extra" testing that needs to happen for a specific driver, something that is driver-specific that won't be in the standard suite, you could easily just add that as an extra test in the same test file.

akutz:
Hi @codenrhoden,

Your view of things is correct. In the end the driver's only test file is like so:

```go
// +build !libstorage_storage_driver libstorage_storage_driver_vfs

package vfs

import (
	"testing"

	apitests "github.com/codedellemc/libstorage/api/tests"

	// load the driver packages
	_ "github.com/codedellemc/libstorage/drivers/storage/vfs/storage"
)

func TestSuite(t *testing.T) {
	apitests.RunSuite(t, "vfs")
}
```

However, there's no reason you could not also execute your own tests from the driver's test file.

akutz:
Hi @dvonthenen,

I'd appreciate if you'd take a look at this PR as well. Please note how the file `vfs_test.go` has changed. It makes writing tests for drivers as simple as calling a single function.

PR #532, opened 2017-05-01
Announce ec2 removal version

codenrhoden:
I'd like to see the backwards compatibility of the EBS driver using "ec2" as the driver name removed for REX-Ray 1.0. The docs have long called it out as deprecated, so I think it's fair to say when we are going to remove it.

This will allow us to remove some code and simplify testing scenarios.

Discussed with @akutz, and he is on board.

Would like to hear thoughts from @cduchesne and @clintonskitson as well.


clintkitson:
+1

PR #536, opened 2017-05-02
Cinder storage driver

akutz:
This patch introduces support for the OpenStack storage driver. This feature's authors include:

* Mathieu Velten <mathieu.velten@cern.ch>
* Joe Topjian <joe@topjian.net>

This PR replaces #243.


akutz:
Hi @cduchesne and @codenrhoden,

This PR has squashed the commits from #243. I left a separate commit, the one that renames this to Cinder, until we all agree to do so. I know @codenrhoden is in favor of it. I agree that if it is ever to happen it should happen now.

This is pending @codenrhoden's testing of the squashed, renamed driver.

PR #541, opened 2017-05-02
Disable TLS for UNIX sockets

akutz:
This patch explicitly disables TLS for libStorage as a client or server when connecting to or serving on a UNIX socket.

/cc @koensayr @cduchesne @codenrhoden

### DIFF
--- a/api/server/server_http.go
+++ b/api/server/server_http.go
@@ -9,6 +9,7 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"strings"
 	"sync"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


codenrhoden (approved):
If we only need to catch this when calling `ParseTLSConfig`, this seems to do it. I grepped the code and this is the only two places it gets called.

akutz:
Hi @codenrhoden,

Yep, it's the only two places. The reason it's a utility method is to handle the possible, multiple roots. That's to say

akutz:
Hi @codenrhoden,

Yep, it's the only two places. The reason it's a utility method is to handle the possible, multiple roots. That's to say it supports:

akutz:
Hi @codenrhoden,

Yep, it's the only two places. The reason it's a utility method is to handle the possible, multiple roots. That's to say it supports:

```yaml
libstorage:
  tls:

**Comment by akutz** [comment]
Hi @codenrhoden,

Yep, it's the only two places. The reason it's a utility method is to handle the possible, multiple roots. That's to say it supports:

```yaml
libstorage:
  tls:
```

*or*

```yaml
libstorage:
  client:
    tls:
```

Because technically the latter should be:

```yaml
libstorage:
  client:
    libstorage:
      tls:
```

For both the `tls` and `logging` properties I made them possible to not have to redefine the entire property path directly under `libstorage.client` and `libstorage.server`.

PR #543, opened 2017-05-03
Integration driver mount retry logic

akutz:
This patch introduces the ability to configure a retry count and duration for integration driver mount operations. It satisfies a condition encountered by a user with the Ceph driver in concert with K8s.

/cc @cduchesne

### DIFF
--- a/api/registry/registry_integration.go
+++ b/api/registry/registry_integration.go
@@ -2,6 +2,7 @@ package registry
 
 import (
 	"sync"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	gofig "github.com/akutz/gofig/types"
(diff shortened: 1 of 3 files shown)


api/registry/registry_integration.go
```diff
@@ -143,10 +154,22 @@ func (d *idm) Mount(
...
+		ctx, volumeID, volumeName, opts)
 	if err != nil {
-		return "", nil, err
+		for x := 0; x < d.retryCount; x++ {
+			mp, vol, err = d.IntegrationDriver.Mount(
+				ctx, volumeID, volumeName, opts)
+			if err == nil {
+				break
+			}
+			time.Sleep(d.retryWait)
```
codenrhoden: My only comment is whether this `Sleep()` should instead be before the  call to `Mount` inside this `for` loop. That way there will be delay between the initial mount attempt and the first retry. As written, if the retry wait was 1, there would be no delay between the initial attempt and the one and only retry.

akutz: You, sir, are correct. Fixing now.

akutz: Fixed.

codenrhoden (approved):

PR #545, opened 2017-05-04
Remove client driver

akutz:
This patch removes a vestigial piece of the framework, the client driver.


akutz:
Hi @codenrhoden,

I've added you as a reviewer on this PR, but I am not necessarily awaiting your review. I've never really taken the time to explain the client driver to you, and it's an unused piece of functionality. Consider that it was originally conceived by me as a way to do what the executor does, but sans a separate binary. We elected not to use this as even though the rule was nothing but stdlib (for keeping the footprint light), the optics of seeing `storage/ebs/client` or whatever in the client's dependency graph wasn't good.

The only thing that's ever used this is the VFS driver, and only for testing purposes. I've relocated those pieces to the VFS's remote side and so the tests continue to pass. I wanted you to be aware of this as it's at first blush non-trivial refactor, and maybe it is, but nothing uses it. So...be aware I'm removing this...can you see any reason I shouldn't or any problem with how I have?

codenrhoden:
Thanks for the history!

PR #546, opened 2017-05-05
Enable TLS for UNIX Sockets via Env Var

akutz:
This patch provides a way to reenable TLS for UNIX sockets by setting the environment variable `LIBSTORAGE_TLS_SOCKITTOME` to a truthy value.


codenrhoden (approved):
Really appreciate the great explanation in the docs of what's going on, and why.

The code changes look good.

PR #557, opened 2017-05-23
Go 1.8 Plug-in Support

akutz:
This patch introduces support for loading Linux shared object (.so) files as Go 1.8 plug-ins. This feature requires Linux.


akutz:
Replaced by #575.

PR #561, opened 2017-05-26
Introduced support for optimized VolumeInspect by name

### LINKED ISSUE #560: Add storage driver option to inspect volume by name
# Summary
I propose to add an optional enhancement to storage drivers, where they can add a `VolumeInspectByName` method that can quickly lookup a single volume on the backend for providers that support looking up a volume by name instead of ID.

# New Feature
There are several motivating factors for this one, so hear me out.

Right now, the storage drivers have `Volumes` and `VolumeInspect` methods. The former returns all volumes it can find on the backend, and the latter returns a single volume, as referenced by it's `volumeID`. For every method that acts on an existing volume, the storage drivers expect a `volumeID` as their unique identifier to the volume in question. When creating a new volume, a name is required, and the `volumeID` is figured out after the volume exists.

From a libStorage client perspective, we see that we spend significant effort figuring out what the right `volumeID` is, because in practice we are usually given a `name` instead of `volumeID`. REX-Ray goes to great lengths to make sure that you can reference a volume by either an ID or a name.

As currently implemented, I think we are wasting a lot of cycles on this. Many of the storage providers allow an API query to "get" a volume by name. What we are doing now is that whenever we need to find the `volumeID`, we first call to `Volumes()`, getting  a list of all volumes, then do a string match on the name to find the right volume. Then we can use the associated `volumeID` for subsequent actions like `attach`, `detach`, etc.  We can see this in REX-Ray's CLI, where something like `rm` first [does an ls](https://github.com/codedellemc/rexray/blob/master/cli/cli/cmds_10_volume.go#L235) to get all volumes and matches the name, then [removes based on the ID](https://github.com/codedellemc/rexray/blob/master/cli/cli/cmds_10_volume.go#L264).

I think it would be better if these methods instead used a `VolumeInspect`, to get the one and only matching Volume. But today, that is not possible because users usually pass in the volume name, not the ID. Furthermore, the Docker volume driver *only* uses names -- it never looks at IDs. Which is why the integration driver [duplicates](https://github.com/codedellemc/libstorage/blob/master/drivers/integration/linux/linux.go#L132-L139) this "volume ls, match on name to get ID" logic that the REX-Ray CLI uses. I realize that REX-Ray has a pattern matching system in place that requires on the `ls` functionality, and I think that can be maintained. I can go into more detail on that later in a REX-Ray issue if we get that far.

I think we can be a lot more efficient by supporting a `VolumeInspect` method in libStorage storage services that accepts a volume name instead of an ID. The libStorage framework actually has a way to do this "match by name" all on the server side today, by setting the `byName` querystring for `VolumeInspect`, as seen [here](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L301-L338). This is an improvement in that it vastly reduces the amount of data transferred back to the caller down to just the matching volume instead of potentially dozens or hundreds, but it is still a heavier call than may be required for many storage providers.

My last reason for wanting this feature is to solve #450 and #458, where we want to automatically replace underscores in a volume name to dashes. It is trivial to do this upon volume creation, but after that point the volume is essentially unaccessible unless you are using volume IDs.  If Docker creates a volume named `test_test`, for example, and then we create `test-test` on the backend, we will always report the volume's name as `test-test` in the future. We can **not transform the volume name back** to `test_test` in queries, as this runs the danger of modifying names that were originally created with dashes on purpose. If there was a `VolumeInspectByName` function for the driver, it could take the name `test_test` and transform it to `test-test` and find the right volume, at which point the caller would have the correct `volumeID`. Without such a function, there is no way for a libStorage client to lookup such a volume by name.

# Proposal
I suggest we add a new interface, something like `StorageDriverWithNameInspect`, and when a `VolumeInspect` request comes into the volume router, we check if the storage driver implements that interface and use it in the `byName` [case](https://github.com/codedellemc/libstorage/blob/master/api/server/router/volume/volume_routes.go#L301-L338) instead of the current methodology. We fall back to the current stuff if the driver doesn't support it.

With that enhancement in place, we then modify the integration driver to use the `byName` flag in combination with `VolumeInspect` instead of the [current `Inspect`](https://github.com/codedellemc/libstorage/blob/master/drivers/integration/linux/linux.go#L132-L139) approach. Keep in mind this already accepts a *name*, not an ID. We would also enhance the `volumeInspectByIDOrName` [logic](https://github.com/codedellemc/libstorage/blob/master/drivers/integration/linux/linux_utils.go#L56-L71) to be more efficient in the "or name" case.

And finally, the end game would be to enhance the REX-Ray CLI to use `VolumeInspect` more appropriately. If this proposal is acted upon, a separate issue would cover the REX-Ray side of things, but I think there are several improvements that can be made there as well. No point in discussing them at length here, as it is all moot unless this libStorage proposal makes sense.

codenrhoden:
This patch introduces support for `InspectVolumeByName` to the libStorage client. This is not a new API, but uses an existing capability of the API that was never consumed by the client. When invoked, the libStorage server will execute an optimized inspection of a single volume based on its name for drivers that support it. This is an enhancement over the existing behavior of the client requesting all volumes, then matching on the name to get the volumeID and calling InspectVolume on that ID.

This does not update any drivers to take advantage of the new support, and therefore should not change any existing behavior.

Fixes #560


codenrhoden:
I've tested this patch with an existing, unmodified driver (dobs) and didn't observe any ill-effects. But still, I'd like to test with a modified driver (that adds `VolumeInspectByName` to see that this patch is correct before saying it is ready for merging.

But it is definitely ready for review.

codenrhoden:
I tried this out with the [`gcepd`](https://github.com/codenrhoden/libstorage/commit/458e305ce1c8986663d3a82908856b05050a9f29) driver, and it worked great!

PR #576, opened 2017-06-13
Allow disabling of RBD modprobe

codenrhoden:
When RBD driver is running in a container, using `modprobe` to check if
the RBD module is already inserted is not apprpriate. Keep that
behavior as the default, but add a config option to disable it that can
be used for container deployments.


codenrhoden:
I *think* this is the only driver modification necessary to support codedellemc/rexray#818.

PR #580, opened 2017-06-21
GCE translate underscores

### LINKED ISSUE #458: Transparently translate underscores to dashes for GCE
@sterling312 commented on [Tue Apr 05 2016](https://github.com/codedellemc/rexray/issues/349)

When creating a rexray volume with docker-compose, the compose will automatically add project name in front of the volume name (default to the folder name of where the compose file is) separated with underscore. 

GCE volumes does not accept underscore as valid volume name. Most likely, rexray gce driver will need to transpose any _ to something safe like - when creating the actual volume name on GCE.

codenrhoden:
This is a bit experimental, but worked completely in my testing with Docker.

New config option, gcepd.convertUnderscores, enables the GCEPD driver to
replace any underscores with dashes during volume creation. When a
VolumeInspectByName request comes in, that name is also converted, and
the volume that was created with the dashes is returned.

It is important to note that the returned data from the driver is always
correct -- the name is never displayed as having underscores instead of
dashes, because no such volume exists.

Fixes #458

without patch:
```
$ docker volume create -d rexray test_me3 --opt size=10
Error response from daemon: create test_me3: VolumeDriver.Create: {"Error":"Volume name does not meet GCE naming requirements"}
```

with patch:
```
$ docker volume create -d rexray test_me4 --opt size=10
test_me4
$ ./rexray -s gcepd volume ls
ID          Name        Status     Size
test-me4    test-me4    available  10
$ docker volume ls
DRIVER              VOLUME NAME
rexray              test-me4
$ docker volume inspect test_me4
[
    {
        "Driver": "rexray",
        "Labels": {},
        "Mountpoint": "/",
        "Name": "test-me4",
        "Options": {
            "size": "10"
        },
        "Scope": "global",
        "Status": {
            "availabilityZone": "us-west1-b",
            "fields": null,
            "iops": 0,
            "name": "test-me4",
            "server": "gcepd",
            "service": "gcepd",
            "size": 10,
            "type": "pd-ssd"
        }
    }
]
$ docker run --rm --volume-driver=rexray -v test_me4:/data busybox mount | grep \/data
/dev/disk/by-id/google-test-me4 on /data type ext4 (rw,seclabel,relatime,data=ordered)
# docker volume rm test_me4
test_me4
```

So, at this point for docker, `test_me4` and `test-me4` are pretty much aliases for each other. Referencing either one works.

This PR depends on #579, and includes the commit from that PR. Once that PR is merged, this can be rebased to remove the additional commit.

### DIFF
--- a/.docs/user-guide/storage-providers.md
+++ b/.docs/user-guide/storage-providers.md
@@ -1013,6 +1013,7 @@ gcepd:
   statusMaxAttempts:  10
   statusInitialDelay: 100ms
   statusTimeout:      2m
+  convertUnderscores: false
 ```
 
 ##### Configuration Notes
@@ -1050,6 +1051,12 @@ gcepd:
 * `statusTimeout` is a maximum length of time that polling for volume status can
   occur. This serves as a backstop against a stuck request of malfunctioning API
   that never returns.
+* `convertUnderscores` is a boolean flag that controls whether the driver will
+  automatically convert underscores to dashes during a volume create request.
+  GCE does not allow underscores in the volume name, but some container
+  orchestrators (e.g. Docker Swarm) automatically prefix volume names
+  with a string containing a dash. This flag enables such requests to proceed,
+  but with the volume name modified.
 
 #### Runtime behavior
 * The GCEPD driver enforces the GCE requirements for disk sizing and naming.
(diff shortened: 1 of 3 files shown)


akutz:
Hi @codenrhoden,

Can you rebase this PR? It has two commits on it. For some reason f30c7cf34a0f9abd02ec8898136d1ad51debd21d isn't going away, even though #579 has been merged (and it has a single commit...the aforementioned f30c7cf34a0f9abd02ec8898136d1ad51debd21d). I don't want to accidentally screw anything up. Please rebase this, and then we can merge it cleanly.

PR #585, opened 2017-07-17
Implement VolumeInspectByName for RBD storage.

vstoican:
Instead of retrieve the information about all the RBDs, individual RBD can be queried. This speeds up storage mounting via Docker rexray. 
This is possible due to https://github.com/codedellemc/libstorage/pull/561


codenrhoden (approved):
Thanks @vstoican! It had been on my list to do this exact same thing.

akutz:
> ```
> // volumeName and volumeID are the same for RBD
> ```

I have a question, if the inspect per RBD is the same for name or ID then why add the ByName variant? Is it to just make it so the RR CLI doesn't have to do the list prior to a direct inspect?

vstoican:
@akutz VolumeInspect @ volume_routes.go invokes either VolumeInspectByName or Volumes (if VolumeInspectByName is not implemented). Volumes queries information about all volumes whilst VolumeInspectByName inspects only the RBD which is being mapped. For a small number of RBDs the there might not be a huge difference. However, for a larger number of RBDs the execution time increases exponentially.

vstoican:
@codenrhoden It was good to help. I think it should be implemented for all the other drivers as well.

akutz:
Hi @vstoican,

Thank you for the explanation. I wrote all that code, but it's been a while. However, what confused me is that the workflow logic is only optimized if the "byName" QS is provided forcing the inspect to look up a volume by its name. I just wanted to be clear that unless the API consumer is doing that this change won't have any effect.

codenrhoden:
@akutz -- just to follow-up, the history here is that the API consumer in this case is the Docker Volume Driver -- which uses the Linux integration driver.

I added support for that here: https://github.com/codedellemc/libstorage/blob/master/drivers/integration/linux/linux_utils.go#L43-L44, in this PR: https://github.com/codedellemc/libstorage/pull/561

So any driver that can implement this can save a lot of time for large number of volumes when using Docker. This enhancement is *not* yet able to be taken advantage of when using the REX-Ray CLI.

@vstoican One interesting point, is that not all storage provider APIs have a way to look up a volume by name. So not every driver can do it.

PR #587, opened 2017-07-18
Package-Level Makefiles for Testing

akutz:
This patch introduces package-level Makefiles for driving tests for the libStorage framework as well as the storage drivers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -713,73 +713,6 @@ GO_CLEAN += $$(PKG_A_$1)-clean
 endif
 endif
 
-
-################################################################################
-##                               PROJECT TESTS                                ##
-################################################################################
-ifneq (,$$(strip $$(TEST_SRCS_$1)))
-ifneq (1,$$(TEST_C_$1))
-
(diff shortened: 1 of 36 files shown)


codenrhoden (commented):
Is there any way to not duplicate all the package level Makefiles? a symlink to a master version or something? I've never committed a symlink to git, though I know it's possible, but I'm not sure of the ramifications.

I do like all the latest changes. will make the project easier to work on by all, and more approachable.

akutz:
Hi @codenrhoden,

A symlink is possible and so is placing a file at the top level called `driver-tests.mk` or `framework-tests.mk` and then in the individual directories including a `Makefile` that has a single line:

```Makefile
include ../../../driver-tests.mk
```

Or something like that.

The reason I didn't opt for referencing a file at the root of the project is for portability. We plan to break the drivers out into other repos, and so I didn't want to link to anything at the root of this one. Does that make sense?

codenrhoden (approved):
More simplifications! I love it.

I know that shows I don't understand go coverage all that well, but why isn't `TEST_COVERPKG` necessary anymore?

PR #593, opened 2017-07-21
Endpoint/Service Init Order Fix

akutz:
This patch fixes an issue where service init failures would cause libStorage endpoints that use UNIX sock files to leave those sock files in place instead of removing them on error. This fix causes the service init routine to now occur *before* the endpoint init routine, thereby causing any service failure to occur before an endpoint's sock file is even created.

There is no issue filed for this fix, but it was a _911_ issue raised by @kacole2 from the REX-Ray Slack channel. @cduchesne reproduced the error [with a log file](https://gist.github.com/cduchesne/3f05a69192b179debd40d81b9053a72c).

### DIFF
--- a/api/server/server.go
+++ b/api/server/server.go
@@ -164,15 +164,15 @@ func newServer(goCtx gocontext.Context, config gofig.Config) (*server, error) {
 
 	s.ctx.Info("initializing server")
 
-	if err := s.initEndpoints(s.ctx); err != nil {
-		return nil, err
-	}
-	s.ctx.Info("initialized endpoints")
-
 	if err := services.Init(s.ctx, s.config); err != nil {
 		return nil, err
 	}
 	s.ctx.Info("initialized services")
+
+	if err := s.initEndpoints(s.ctx); err != nil {
+		return nil, err
+	}
+	s.ctx.Info("initialized endpoints")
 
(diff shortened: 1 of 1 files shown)


codenrhoden (approved):
Looks good!

So, this issue begs the question (again) of whether we need to cut another minor release. *IF* we decide to, there are other things I'd like to get in. And *IF* we decide to, would you rather target this at a `release/0.6.3` branch, such that we can merge it to master post release and not have to worry about cherry-picking?  I know there are some major changes that have gone into libStorage since the last release, and don't think we'd want to build off of those for a new release.

I really don't know if we want to do another one or not, but if not, we're holding this fix until September.

akutz:
Hi @codenrhoden,

Once @cduchesne verifies it I'm going to merge this to master. At that point I'd have very few qualms about just relocating libStorage as a whole into `github.com/codedellemc/rexray/pkg/libstorage` and simply rewriting the paths using `gofmt`'s package rewrite function. If we want to tag libStorage `master` as `0.6.3` so we know exactly which bits we copied over, then I'm fine with that too.

akutz:
Hi @codenrhoden,

I guess my overall point is that if we collapse LS into RR sooner rather than later then the decision to do a minor release is only about RR instead of about both of them.

codenrhoden:
@akutz roger that. 👍

PR #596, opened 2017-07-26
Update README

akutz:
This patch updates the README to reflect the relocation of libStorage. Please use [this link](https://github.com/akutz/libstorage/blob/feature/update-readme/README.md) to view the formatted README.


codenrhoden (approved):
Thumbs up from me. I feel like it's right in line with many other projects I've seen that get deprecated/transferred/EOL'd.

I think we'll want to leave this repo up indefinitely, so having this direct link to where the action is at now is important.

clintkitson:
@akutz there are two things I'm thinking about here.. 1) should we include plan for 1st release of RR at 0.10 to include this merged libstorage code 2) should we reference CSI's impact, ie. making libstorage less important for projects like RR?

codenrhoden:
Can we come to a consensus on the new README? I think it's bad that we still don't reflect the current state of this repo. Even if we want to enhance later, I think it's better to reflect what is going on now and tell me people libStorage has moved into REX-Ray.

akutz:
This is on me. I will update this with Clint's suggestions and merge this.

